PR #203, opened 2015-06-21
scale Picons in ChanneSelection

lolly-enigma2:
If Picons have great Diminsions (420x240) will it no visible in
ChanneSelection with old code.

Now with scaling the picons are visible in ChanneSelection


lolly-enigma2:
That was great bullshit ;)

PR #219, opened 2015-07-31
record AITs

philotas:
I had the use case where I needed the AITs inside the TS for a recording.
I made the changes as suggested here: http://forums.openpli.org/topic/38513-ait-cropped-from-transport-stream/ and it seems to work fine.

### DIFF
--- a/lib/service/servicedvbrecord.cpp
+++ b/lib/service/servicedvbrecord.cpp
@@ -436,8 +436,10 @@ int eDVBServiceRecord::doRecord()
 					if (i->capid >= 0) pids_to_record.insert(i->capid);
 				}
 			}
-
-				/* find out which pids are NEW and which pids are obsolete.. */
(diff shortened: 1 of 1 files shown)


philotas:
I can see that the use case only applies to very few users, so it could maybe be an option which is disabled by default although it should do no harm having it activated by default.

arn354:
I added it with a recording config option.
https://github.com/openatv/enigma2/commit/6d2101cc3c6dbd6850868993f0cdece3f88d347c

Schimmelreiter:
What would be the consequence if we just enable it w/o option to switch it off?
I think E2 already has too many confusing options.

arn354:
PLi added it without config option. Personally i prefer to be able to configure anything "additional". But i have basically no obligations if you want to change it and enable it as default. Additional AIT data in recordings seem not to be that much.

PR #269, opened 2015-11-15
Update it.po

CloudItaly:


atvcaptain:
please check it.po beforre commit we have many typos in the file  check with msgfmt -o it.mo it.po

PR #300, opened 2016-03-15
Event description scrollable in SecondInfoBar

pizzelnet:
Makes event description scrollable in SecondInfoBar (screen)


pizzelnet:
To make event description scrollable in SecondInfoBar (screen).

PR #310, opened 2016-03-21
Add Maximum number of days in EPG

pizzelnet:
Cannot check, sorry.


atvcaptain:

PR #313, opened 2016-03-21
Update setup.xml

pizzelnet:


atvcaptain:

PR #317, opened 2016-03-24
File size in GB with 2 digits when file is +1024MB

pizzelnet:
Shows the size of the file in gigabytes (with two digits for megabytes) when it is higher than 1024 MB. (E.g. 4.35GB)


pizzelnet:
gigabyte

pizzelnet:
gigabyte

PR #323, opened 2016-03-25
Update epgcache.cpp

pizzelnet:


atvcaptain:
you break a function with this commit

bool save_epg = eConfigManager::getConfigBoolValue("config.epg.saveepg");

this need then epg.dat save to nas and more then one box share the same epg dat

PR #330, opened 2016-04-02
Updated the PO Translation files.

neoatomic:
I noted that many of the PO files where not up-to-date and didn't contain the latest entry's from the POT file. So i ran them all trough Poedit.

* Added the latest entry's from the POT file.
* Validated all files with Poedit, so they should be error free.
* Poedit made the PO's a little more uniform.
* Changed the Project IDs to `OpenATV`.
* The `ku` Kurdish file was set to `en` (English) fixed that also.

There are a few PO Language files that don't have the recommended `Plural-Forms` setting according to Poedit.
i'm not sure why that is, so i didn't touch/change them. But those are `AR/CS/FR/HR/LV/RO/SL/TH/TR/`

As this is a large PR, please test/review independently.


madieatv:
This branch has conflicts that must be resolved

PR #352, opened 2016-05-12
Fix epgcache.h

pizzelnet:


atvcaptain:
looks 100% identical ?

Schimmelreiter:
The bash'ish line break is gone ...

PR #407, opened 2016-07-06
Greek translation update 6/7/2016

MCelliotG:
Full Greek translation update in sync with enigma2.pot and de.po

### DIFF
--- a/po/el.po
+++ b/po/el.po
@@ -2,7 +2,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: OpenATV Enigma2 Official Greek Translation\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


atvcaptain:
cant merge not in sync !

MCelliotG:
Let me check it and I'll try again. Perhaps because the de.po file contains about 1000 strings that are not included in the enigma2.pot! Shouldn't these strings be available as an addition for all other languages as well?

atvcaptain:
we must create new pot file , but i have less time for this, we planing in the feature a better solution platform for translation, but need time, the german po file is complete other po files some parts missing

MCelliotG:
I can take care of creating a new pot if you like me to.

PR #424, opened 2016-07-31
Add option of re-assigning coloured key within G-EPG

bbbuk:
This adds option of allowing (to a certain degree) re-assigning of the four coloured keys buttons within Graphical EPG.

Thanks to bbbuk/ketmp/darkdreamdan/matt007


atvcaptain:
i test first then ok i merge the parts

hd75hd:
this commit brokes the behavior of the green button when we scroll in the simple EPG ( also other EPGs ) over a event with a record-timer

bbbuk:
I will try to look at that over weekend.  Most of the code has if statements to determine that the current epg in view is graphical one before.

PR #433, opened 2016-08-11
epgcache.cpp: do not save cache on MHW2

djMesias:
Save mhw(new)-epg after update can cause segfault.
Cache must be saved on exit or by external call.

It solves segfault when try to update mhw-epg from SID 30150
and EPG is set to 4 or more days.


pizzelnet:
@djMesias how do you do to save the EPG for mhw(new)?

PR #441, opened 2016-08-22
add LAN/WLAN IP  in  about screen

undertaker01:


atvcaptain:
we have the info in about network so we dont need this

PR #449, opened 2016-08-29
add delete to load/save in epg menu

davesayers2014:


davesayers2014:
Thanks there maybe more once I see them.

davesayers2014:
It functions but I now see what I missed 

+class EpgDeleteMsg(MessageBox):
 +	def __init__(self, session):
 +		MessageBox.__init__(self, session, _("Are you sure you want to delete the EPG data from:\n") + config.misc.epgcache_filename.value, MessageBox.TYPE_YESNO, default=False)
 +		self.skinName = "MessageBox"

this info is in the load/save.py class EpgDeleteMsg(MessageBox):
	def __init__(self, session):
		MessageBox.__init__(self, session, _("Are you sure you want to delete the EPG data from:\n") + config.misc.epgcache_filename.value, MessageBox.TYPE_YESNO, default=False)
		self.skinName = "MessageBox"

https://github.com/openatv/enigma2/blob/master/lib/python/Components/EpgLoadSave.py

PR #485, opened 2016-10-16
Add NextEvents - thx tecfan69

arn354:
see http://www.opena.tv/die-bastelecke/28812-chennelselection-verbesserungsvorschlag-post254914.html#post254914


arn354:
needed to revert as config.epgselection.overjump creates issues i am not able to solve quickly.

PR #622, opened 2017-02-13
Missing colorlcd800 entry in enigma2.bb

neoatomic:
Currently the LCD skin for the ultimo4k (800x480) is not shown on the feed for download/install.


atvcaptain:
the enigma2.bb is only for debugging not use for normal build, but ok we merge it for feature use

neoatomic:
Thanks.
Currently the new 800 lcd skin is not shown on the feed for the ultimo4k. So i went trough the build process but could not find any errors... Except this missing entry. So thought this might be the problem. ;)

When you find time, could you do a quick check why the skin isn't shown on the feed.? I wasn't able to find the problem. ;(

atvcaptain:
https://github.com/oe-alliance/oe-alliance-core/commit/59b9dc4bad54d6f22e683dc6cc3ac777c1e0edae

PR #697, opened 2017-05-27
Zgemmah3 use same way like h7

a4tech:


BliNeR-KeY:
ohhh thanks finally to remove VFDCONTROL.
just maked low of decoders for lot of time dhis system plugin.

thank you atv,egami,taapat  again.

hope fast dhis patch in others 7seg decoders.

PR #700, opened 2017-06-06
[SystemInfo] add 'gbquad4k' to HDMIin

madieatv:


atvcaptain:
?? gb7252 allready include  gqquad4k was boxtype

PR #703, opened 2017-06-10
[Flashonline] prepare for image 6.1

madieatv:


Schimmelreiter:
We should somehow automate this ...
e.g. by downloading a file which specifies which version is current stable and which one is current beta.

Issue #759, opened 2017-08-24
NimManager


AbuBaniaz:
Can someone please look at adding this to OpenATV please?
https://github.com/OpenViX/enigma2/commit/0d09a41b346f94f30e2619c571c28fe2b97b7a14#diff-021996494dde61d73519d0ab55099cb4

I tried, but could not resolve the merge conflicts.

At the moment OpenATV still saves the position number within the cables.xml and terrestrial.xml. When a new entry is added it alters things. Yes, people can say that add to the bottom. But that just makes it awkward having entries all over the place.

A few images have adopted this way of doing things already. PLi did it here
https://github.com/OpenPLi/enigma2/pull/1050

Thanks in advance.

Other picks for you to consider...

https://github.com/OpenPLi/enigma2/pull/1054/commits/ebb1502b48364ef3b29c081505ed6c70f38b3c06

https://github.com/OpenPLi/enigma2/pull/1057/commits/d4f6562f3544cfb3e27d79402e7a72e521c52140

PR #761, opened 2017-08-27
[NimManager] Use name in xml file, not position number.

AbuBaniaz:
Deal with issue #759 

At the moment, the entry in settings file uses the position number within the xml file.  As an example, settings file contains ```config.Nims.1.dvbt.terrestrial=70```
If the order changes or another provider is added, the position number obviously changes and users are unwittingly using a different provider.

This is what the pull request will achieve.
```
config.Nims.1.dvbt.terrestrial=Crystal Palace, London (Europe DVB-T/T2)
```

There will be a need to re-save cable/terrestrial config after this commit. Small loss for future robustness.

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -2021,10 +2021,10 @@ def InitNimManager(nimmgr, update_slots = []):
 		try:
 			nim.scan_networkid
 		except:
-			list = [(str(n), x[0]) for n, x in enumerate(nimmgr.cablesList)]
+			list = [(x[0], x[0]) for x in nimmgr.cablesList]
 			nim.scan_networkid = ConfigInteger(default = 0, limits = (0, 99999))
 			possible_scan_types = [("bands", _("Frequency bands")), ("steps", _("Frequency steps"))]
-			if n:
+			if list:
 				possible_scan_types.append(("provider", _("Provider")))
 				nim.scan_provider = ConfigSelection(default = "0", choices = list)
 			nim.scan_type = ConfigSelection(default = "provider", choices = possible_scan_types)
@@ -2055,7 +2055,7 @@ def InitNimManager(nimmgr, update_slots = []):
 		try:
 			nim.terrestrial
 		except:
-			list = [(str(n), x[0]) for n, x in enumerate(nimmgr.terrestrialsList)]
+			list = [(x[0], x[0]) for x in nimmgr.terrestrialsList]
 			nim.terrestrial = ConfigSelection(choices = list)
 			nim.terrestrial_5V = ConfigOnOff()
 
@@ -2063,7 +2063,7 @@ def InitNimManager(nimmgr, update_slots = []):
 		try:
 			nim.atsc
 		except:
-			list = [(str(n), x[0]) for n, x in enumerate(nimmgr.atscList)]
+			list = [(x[0], x[0]) for x in nimmgr.atscList]
 			nim.atsc = ConfigSelection(choices = list)
 
 	try:
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -2021,7 +2021,7 @@ def InitNimManager(nimmgr, update_slots = []):
 		try:
 			nim.scan_networkid
 		except:
-			list = [(str(n), x[0]) for n, x in enumerate(nimmgr.cablesList)]
+			list = [(x[0], x[0]) for x in nimmgr.cablesList]
 			nim.scan_networkid = ConfigInteger(default = 0, limits = (0, 99999))
 			possible_scan_types = [("bands", _("Frequency bands")), ("steps", _("Frequency steps"))]
 			if n:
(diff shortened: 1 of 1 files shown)


Schimmelreiter:
This just shifts the problem, it does not solve it.
The proper solution would have to be an additional unique ID which does not change at all or just keep the number and sort the list on the fly, rather than to notoriously keep it alphabetically ordered.

What we have now:
1 - A cable provider
2 - Other cable provider

User selects "Other cable provider", 2 gets saved.

Someone adds "Another cable provider" alphabetically, users previously on "Other cable provider" are now on "Another cable provider" and need to re-configure their provider to 3.

What we get now:
1 - A cable provider
2 - Another cable provider
2 - Other cable provider
User selects "Other cable provider", "Other cable provider" gets saved.
"Other cable provider" gets bought by "A cable provider" and networks get merged, someone removes "Other cable provider" from the list and all users previously on "Other cable provider" are now on an illegal value.

On the other hand, renames have never and will never occur ...
.... oh, ooops:
ish -> ish/iesy -> Unitymedia
iesy -> ish/iesy -> Unitymedia
KabelBW -> Unitymedia
Kabel Deutschland -> Vodafone

Multikabel -> Ziggo
@Home Network -> Ziggo
Casema -> Ziggo
UPC Nederland -> Ziggo

(Initially) one unique ID per provider is the only real thing, where a provider may get multiple IDs through mergers in the long run.

AbuBaniaz:
The pull request will save by name, not position number. The user selection will not change so long as the name does not change no matter what happens above it in the list. It is a step in the right direction until you.

Where are these unique provider IDs you talk about? We can start using them as soon as you provide them.

In any case, when you add these provider IDs, the tuner config will need to save by the actual ID, not the position number of the ID within the xml file.

Huevos:
> "Other cable provider" from the list and all users previously on "Other cable provider" are now on an illegal value.

No. You are wrong. There is no BSoD if the provider is removed from the list. If a provider is removed from the list the code defaults to brute force cable scan, or in the case of terrestrial, default Europe values. This code has been used in OpenViX for a long time, and in OpenPLi and OpenMIPS for several months without a single problem reported.

Huevos:
And using the list index or unique IDs just makes the settings file unreadable for human beings.

Schimmelreiter:
I know what the pull request does.
It's a shift of the problem from "If someone adds providers before ..." to "If the provider gets renamed".

Unique IDs would be something we would have to agree an, like
"CC-SSSSS"
where CC is the country code and SSSSS is a short name for the provider.

E.g.
DE-UMKBW
DE-VFKD
NL-ZIGGO

Even if we later change the **name** of "Unitymedia/KabelBW" to just "Unitymedia" later, the ID "DE-UMKBW" would remain unchanged.

On provider mergers, the remaining one gets the ID of the swallowed one in addition, e.g. if we had done this years ago, "Unitymedia/KabelBW" would now have the IDs "DE-ISH", "DE-IESY" and "DE-KBW".

That would be a proper solution, as it fixes name changes and provider additions at the same time.
It even covers merges, just not splits (But that can't be done anyways, it's actually the same as a provider change).

Huevos:
Which just causes the settings file not to be human-readable and if a provider is deleted we have the same problem with the "unique" IDs.

AbuBaniaz:
No problem, I'll close the pull request as quite clearly you don't want it.

Schimmelreiter:
I just pointed out that is no final solution.

AbuBaniaz:
Your posts portrayed and represented rejection.

There is nothing stopping you or anyone devising a better method. Until that time, we shouldn't really hinder mitigating problems.

Schimmelreiter:
My posts just portrayed my opinion that this is no final solution.
It doesn't matter much if you associate elements by number or by name, as long as **both** can change.
Database 101.
https://code.tutsplus.com/articles/sql-for-beginners-part-3-database-relationships--net-8561

And I just didn't merge this pull because I'm not Mr. OpenATV, there is a team.

Deleting the branch took away the chance for Captain to merge.

Huevos:
It is a just a simple change that sorts out a very common problem (index change). It isn't a purfect solution but name changes are for sure a much less frequent occurrence.

If you are interested in the commit I can send another PR.

Schimmelreiter:
Here is the point:
Switching from index to name requires the user to correct the provider once again.

That's why I'm not in a hurry to merge it, just in case I can come up with a better solution within a foreseeable time period (It would require the user to correct the provider too ...).

But please open another PR, just in case I can't come up with something better soon or Captain deems it fit.

AbuBaniaz:
@Huevos, you might as well as submit the country code stuff too.

Schimmelreiter:
Having seperate fields for cc (country code) and id might be even better ...
A German entry will never move to the netherlands and vice versa :)

Huevos:
We already have country code selection in OpenViX, OpenPLi and OpenMIPS

arn354:
It still can be applied.
The required patch is still there - no need for the remotebranch.
simply add .patch to the pull-request number - it can be downloaded and locally applied - you just don't see it was a Pull-Request/Merge anymore - you just see the commits itself.
https://patch-diff.githubusercontent.com/raw/openatv/enigma2/pull/761.patch

AbuBaniaz:
I've re-submitted here
https://github.com/openatv/enigma2/pull/763

BTW
On OpenVix, we prefer to "squash and merge"

PR #844, opened 2017-11-15
RTSP server improvements

catalinii:
- changes how the read_dmx handles non blocking dmx handle: 
     - if there is a reply that needs to be sent to the server, read_dmx stops reading from the dmx handle and returns to write the data to the client
    - increase the number of TS packets read to 42 to reduce the number of reads call
- increase the buffer in demux.cpp to account for the RTSP header, otherwise, half of the buffers will be just couple of KB
- other changes in filepush.cpp are just indenting changes


catalinii:
And all the changes that @athoik pushed previously (basically changing char * to stringstream)

PR #847, opened 2017-11-19
Update enigma2.pot

thawtes:


Schimmelreiter:
I think you missed some new strings :)

thawtes:
i think so :)

PR #871, opened 2018-01-14
Update de.po

TomS2:
Various orthographical and grammar error corrections, streamlined various terms, created cosistent wordings and put help texts in better context


madieatv:
siehe slack. geht so nicht.

PR #875, opened 2018-01-16
Greek translation update 15/1/2017

MCelliotG:


atvcaptain:
we merge self you must add only to 6.1

PR #987, opened 2018-06-23
multiboot restart/Image Flash - change from STARTUP to slot with empt…

TwolDE2:
…y slot/installed image & build information
Multiboot restart - list installed images with build version/build date, handle multiple bootmode 1/12 for HD51 type receivers.
Flash Online - list slots with empty slot/current installed image
Reduce multiboot entries in SystemInfo to 2 entries
Reduce multiboot restart to only 1 module.
handle different receiver slot numbers and /dev/mmcblkxp.. where x is a receiver variable
Requires: skin change to support changed Flash_Online (new config entry - slot list)
Tested on HD51 and Gigablue 4K's - but not Ceryon & Xcore but code should handle unless "/sys/firmware/devicetree/base/chosen/bootargs" is very different.
Modules changed : SystemInfo, menu.xml,  MultiBootStartup (plus derivatives), Flash_Online, ButtonSetup
Module added: Multiboot.py in Tools


arn354:
i am only able to verify on GigaBlue and hd51 - seems ok.
Other boxes with internal multiboot captain or someone else has to check.

PR #1021, opened 2018-09-12
Yeyo

yeyo:
Polish translation update


yeyo:
I'm really don't get it. I've removed my fork, recreated it, clone to hard disk, created my own branch, replaced pl.po with modified file, commited and regested pull. I'm trying to understand what I'm doing wrong but I don't get it. I've fixed a couple of lines within github gui - is all now correct? I think I have to create my own project first to practice github... (user of svn...)

PR #1030, opened 2018-10-07
few small fixes for automatic audio language selection

### LINKED ISSUE #935: service language transmitted as lang1/lang2
Hi,
I am new to github and enigma2, but I think I found a bug and a solution.

on line 153 in /lib/python/Screens/AudioSelection.py there is a prove:
`i = audio.getTrackInfo(x)
languages = i.getLanguage().split('/')`
 that sometimes a service's audio track language is transmitted as language1/language2 and not just language. For example Slovak / Czech and not just Slovak or Czech.

so the autolanguage setting will not work if I am assuming correctly that the language is selected according to code in /lib/python/Screens/InfoBar.py on line 408:
`lang = audioInfo.getLanguage()
	if langC.has_key(lang):
lang = langC[lang][0]`

if that's the case, then I suppose this should fix it:
`lang = audioInfo.getLanguage().split('/')[0]`

What do you think?

And what is with config.autolanguage.audio_usecache or "Prefer audio track stored by service" in the language settings?
I thought that if I set it to yes it will work like this:
- see if user previously selected some audio track language
- if yes set this language when switching program
- if the user didnt set anything the use the autolanguage1,2,3 or 4 option
should it work like this? because it is not.

Thank you

Vojtech

vojtechj:
This is my first pull request ever.
This fixes #935.

### DIFF
--- a/lib/dvb/pmt.cpp
+++ b/lib/dvb/pmt.cpp
@@ -572,20 +572,21 @@ int eDVBServicePMTHandler::getProgramInfo(program &program)
 		int autoaudio_aache = -1;
 		int autoaudio_aac = -1;
 		int autoaudio_level = 4;
+		int *autoaudio = NULL;
 
 		std::string configvalue;
 		std::vector<std::string> autoaudio_languages;
 		configvalue = eConfigManager::getConfigValue("config.autolanguage.audio_autoselect1");
-		if (configvalue != "" && configvalue != "None")
+		if (configvalue != "" && configvalue != "---")
 			autoaudio_languages.push_back(configvalue);
 		configvalue = eConfigManager::getConfigValue("config.autolanguage.audio_autoselect2");
-		if (configvalue != "" && configvalue != "None")
+		if (configvalue != "" && configvalue != "---")
 			autoaudio_languages.push_back(configvalue);
 		configvalue = eConfigManager::getConfigValue("config.autolanguage.audio_autoselect3");
-		if (configvalue != "" && configvalue != "None")
+		if (configvalue != "" && configvalue != "---")
 			autoaudio_languages.push_back(configvalue);
 		configvalue = eConfigManager::getConfigValue("config.autolanguage.audio_autoselect4");
-		if (configvalue != "" && configvalue != "None")
+		if (configvalue != "" && configvalue != "---")
 			autoaudio_languages.push_back(configvalue);
 
 		int autosub_txt_normal = -1;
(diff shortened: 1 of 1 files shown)


lib/dvb/pmt.cpp
```diff
@@ -572,20 +572,21 @@ int eDVBServicePMTHandler::getProgramInfo(program &program)
 		int autoaudio_aache = -1;
 		int autoaudio_aac = -1;
 		int autoaudio_level = 4;
+		int *autoaudio = NULL;
 
 		std::string configvalue;
 		std::vector<std::string> autoaudio_languages;
 		configvalue = eConfigManager::getConfigValue("config.autolanguage.audio_autoselect1");
-		if (configvalue != "" && configvalue != "None")
```
vojtechj: the real value in config.autolanguage.audio_autoselect* is not "None" but "---".

lib/dvb/pmt.cpp
```diff
@@ -653,18 +654,28 @@ int eDVBServicePMTHandler::getProgramInfo(program &program)
 				int x = 1;
 				for (std::vector<std::string>::iterator it = autoaudio_languages.begin();x <= autoaudio_level && it != autoaudio_languages.end();x++,it++)
 				{
-					if ((*it).find(program.audioStreams[i].language_code) != std::string::npos)
+					if ((*it).find(program.audioStreams[i].language_code.substr(0, program.audioStreams[i].language_code.find('/'))) != std::string::npos)
```
vojtechj: this fixes when language code is transmitted as lang1/lang2.
It could also be fixed in lib/dvb/pmtparse.cpp on line 367, but I can't tell if it's a better solution.
If the line is too long, then the call to substr(0, prog...) can be shortened to substr(0,3), but I think that this tells the reader more about what is happening.

lib/dvb/pmt.cpp
```diff
@@ -715,16 +726,8 @@ int eDVBServicePMTHandler::getProgramInfo(program &program)
...
-				program.defaultAudioStream = autoaudio_ac3;
-			else if (autoaudio_ddp != -1)
-				program.defaultAudioStream = autoaudio_ddp;
-			else if (autoaudio_aache != -1)
-				program.defaultAudioStream = autoaudio_aache;
-			else if (autoaudio_aac != -1)
-				program.defaultAudioStream = autoaudio_aac;
+			if (autoaudio != NULL)
+				program.defaultAudioStream = *autoaudio;
 			else if (first_non_mpeg != -1 && (defaultac3 || defaultddp))
```
vojtechj: using && (defaultac3 || defaultddp) is from my point of view not the best solution in this part of code where everything else (use cache, use ac3, use ac3+, find stream based on languge) failed.
I would suggest adding another option, for example:
default_non_mpeg: Use non mpeg track when no set audio language was found.

vojtechj:
Is there some reason to use integer for some variables in this part of code instead some smaller type (e.g. char) ?
I dont think, that the number of audio streams would ever exceed 127.

PR #1043, opened 2018-11-16
some changes from openpli, and my changes for estring.

nx111:
I  pick some changes from openpli, and add support to enforce remap encode for estring.


atvcaptain:
user report some now broken, please check and fix it

https://www.opena.tv/vu-solo-2-a/43954-6-3-20181117-a.html#post376152

PR #1047, opened 2018-11-22
revert encoding remapping and huffman string convert logic.

nx111:


atvcaptain:
have you fix the issue user have report ?

nx111:
they report that encoded by huffman? if that, may be fixed.

atvcaptain:
we get no feedback from you , have you see this: https://github.com/openatv/enigma2/pull/1043
so we have frist revert all, many user report issue after the big changes, so we must test first more before we can add again

nx111:
I rebase the patches, you can merge after tested it.

PR #1048, opened 2018-11-22
revert encoding remapping and huffman string convert logic.

nx111:
Do not merge first, do it after testing .


atvcaptain:
we get infos from Ukrainian  users :
Hello on the problem of displaying the names of programs in the channel menu on Russian channels
if in /usr/share/enigma2/encoding.conf replace ISO8859-1 with ISO8859-5 the problem disappears

nx111:
I found the reason, and build complete. uploading to google drive. after someone tested it, I will reopen it.

atvcaptain:
we get new request, https://www.opena.tv/fehlende-oder-falsche-uebersetzungen/44029-hope-e2-add-chinese.html   

looks e2 need new front do display add infos correct too

PR #1133, opened 2019-04-16
string format for init.d script

s3n0:
a well-arranged string instead of a list-type, for "init.d" script content


atvcaptain:
we rework all not needed anymore

PR #1146, opened 2019-05-03
Update enigma2.pot

thawtes:
update from reworks ...

### DIFF
--- a/po/enigma2.pot
+++ b/po/enigma2.pot
@@ -1,176 +1,18 @@
-# SOME DESCRIPTIVE TITLE.
-# Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
-# This file is distributed under the same license as the PACKAGE package.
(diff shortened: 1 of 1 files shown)


atvcaptain:
can you check your pot maker we have see its not 100% complete, parts missing in the pot

thawtes:
> 
> 
> can you check your pot maker we have see its not 100% complete, parts missing in the pot

my pot file is 100% correct and compatible!

current pot file string number 6172 bu my pot file strings 6249 (without unnecessary strings)
i am not using pot maker program! because the program cannot fully extract strings!!
I extracted the strings by checking the PY and XML one by one.

I think there is another problem I think there is another reason you do not accept my pull requests.

but it doesn't matter anymore, after that I won't be in every pull request!

good luck ...

atvcaptain:
i dont unterstand, spacerat have start cleanup the de.po working with your pot and many strings missing, we have create own pot and looks better so we must find the problem not more

PR #1183, opened 2019-05-18
Update NumericalTextInput.py

badsystem:


atvcaptain:
we rework soon all, wait after rework

PR #1276, opened 2019-06-11
[VirtualKeyBoard.py] Reorganise image overlay stack

IanSav:
This change moves the character glyph / text / icon to the top layer of the image stack.  Previously the highlight image was at the top of the stack.  This change allows skin designers to have more options in skinning the VirtualKeyBoard with fewer compromises.

### DIFF
--- a/lib/python/Screens/VirtualKeyBoard.py
+++ b/lib/python/Screens/VirtualKeyBoard.py
@@ -962,6 +962,19 @@ class VirtualKeyBoard(Screen, HelpableScreen):
 					w = self.bg_r.size().width()
 					res.append(MultiContentEntryPixmapAlphaBlend(pos=(x, 0), size=(w, self.height), png=self.bg_r))
 					x += w
+				highlight = self.keyHighlights.get(key.upper(), (None, None, None))  # Check if the cell needs to be highlighted.
+				if highlight[0] is None or highlight[1] is None or highlight[2] is None:  # If available display the cell highlight.
+					xHighlight += self.width * width
+				else:
+					w = highlight[0].size().width()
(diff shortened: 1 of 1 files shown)


kiddac:
Ideally the highlight text should be moved below the colour outline. If you have a slightly opague white highlight. It will make reds pink/ blues light blue etc. 

Background... selected highlight... Colour Border. Glyph/Text

IanSav:
Hi Kiddac,

I had the highlight covering the text but I received a complaint that with a semi opaque highlight the text because harder to read.  The solution was to draw the text on top of the highlight layer.

Can you please give me an example of an issue with the current code.  I may be able to find a solution for you.  The solution could be made to work for everyone.

Regards,
Ian.

kiddac:
I have skinned the virtual keyboard very similar to an android device style keyboard.

image

My selected highlight is 20% white.

If I then move the highlight over a keyboard button that has a coloured border. i.e in this example "Exit" The selected highlight goes over the red. Surely it should go under?

image

IanSav:
So when you move over a coloured boundary the boundary is hidden?  The solution is to make the cursor have a transparent edge to allow the border colour show through.  :)

kiddac:
image

kiddac:
> 
> 
> So when you move over a coloured boundary the boundary is obscured? The solution is to make the cursor have a transparent edge to allow the border colour show through without the shading. :)

And you can't really do that, because the highlight will not fit nicely over the other buttons then. Ugly. 
Or the other option is the coloured border is bigger than than the other keys. Even Uglier.

kiddac:
Taking a screenshot of openatv skin, in my opinion that is not right.
The red should always be shown, which like you have done originally, just needs to be reordered.

image

PR #1299, opened 2019-06-23
Update enigma2.pot

badsystem:
add


Schimmelreiter:
I'll update with enigma2.pot generated from build process later.

PR #1301, opened 2019-06-24
Update About.py

badsystem:


atvcaptain:
we dont like 2 tabs  in about

PR #1304, opened 2019-06-24
Update uk.po

badsystem:


Schimmelreiter:
Make sure all translated strings start and end as the original string!
```
msgid "%Y %B %-d"
msgstr "%Y %B %-d "
```

Original string has no trailing whitespace, translated string does.

PR #1307, opened 2019-06-25
about - use GetCurrentImage, GetCurrentImageMode for multiboot capabl…

TwolDE2:
…e receivers to show current STARTUP/slot number
Should also show current bootmode for Zgemma H7/Mutant HD51
Tested on Octagon SF8008 (all 3 slots), Edision Osmio4K and Zgemma h9Combo


nickersk:
can you please adjust conflicting code and reopen PR if needed? thx.

PR #1336, opened 2019-07-14
Update uk.po

badsystem:


Schimmelreiter:
All checks failed:
uk.po:920:15: invalid control sequence
uk.po:924:15: invalid control sequence
uk.po:924:18: invalid control sequence

PR #1338, opened 2019-07-15
[Directories.py] Change resolution path name

IanSav:
This change renames the resolution search directories from "skin_\<resolution\>" to "skin_fallback_\<resolution\>" to avoid confusion expressed by some developers.

The "skin_\<resolution\>" names are still in place to maintain functionality.  They will be removed when the skin repository updates are completed.


IanSav:
The OE-Alliance default skin (skin_default) has already been updated with this path name change.

PR #1372, opened 2019-08-23
[enigma2.sh.in] Revert skin based backdrop logo

IanSav:
This proposed change is not acceptable to all images.  I will revert to the previous mode of operation.


iPkay:
And not return to $datadir instead of $datarootdir (line 58)?

IanSav:
Fixed.  Thank you for spotting the error.

PR #1375, opened 2019-08-26
fix typo NumericalTextInput

thawtes:


IanSav:
Thank you for spotting and reporting this issue.  I will update all the other copies as well.

IanSav:
Pull requests with this same fix now raised for OpenPLi and OpenViX.

PR #1511, opened 2020-01-26
[Keyboard.py] Refactor the code

IanSav:
- Obtain location of keymaps via SCOPE_KEYMAPS.
- Use Pythonic style for processing the "*.info" files.
- Report errors processing the "*.info" files.
- Simplify and optimise the keymap file parsing.
- Log the available keymaps.
- Document a proposal for finding language appropriate default keymap.

### DIFF
--- a/lib/python/Components/Keyboard.py
+++ b/lib/python/Components/Keyboard.py
@@ -1,52 +1,63 @@
+from os import listdir
+
 from Components.Console import Console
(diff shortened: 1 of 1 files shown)


IanSav:
By the way, Australia uses the US keyboard layout!

As this is a refactor I left the code as is but I suggest that it could / should be fixed.

PR #1555, opened 2020-03-14
[Timezones.py] Add error protection to decode()

IanSav:


nickersk:
@IanSav users using vpn are getting wrong tz settings. Any idea how to avoid this ? (like detecting if tun interface exists, and if so, catch user selected tz, etc.)

IanSav:
Once the users set their correct timezone then the geolocation value will no longer be used.

IanSav:
By the way, does OpenATV want to use the new UI for setting the timezone?  Have a look at the code in OpenPLi or OpenViX for the simple changes that need to be made.  Alternatively I can make the changes for you.

nickersk:
user using vpn complained that he always gets the wrong tz, no matter that he setted tz manually...

IanSav:
Can you please get their log files?

nickersk:
13:31:09.9768 {   } Components/Timezones.py:64 InitTimeZones [Timezones] Classic mode with geolocation tz='America/Toronto', area='Classic', zone='America/Toronto'.
13:31:09.9783 {   } Components/Timezones.py:255 activateTimezone [Timezones] Setting timezone to 'America/Toronto'.
quotation: "uhrzeit ist toronto ,da nach neustart hier immer immer toronto als Uhrzeit angezeigt wird hd61,obwohl ich auf berlin umstelle und nach gui neustart ist wieder toronto da"

IanSav:
English please.

I assume they are not in Toronto.

nickersk:
simplified trans: after manual setting tz to berlin, after gui restart there is toronto again

IanSav:
I will need to look at this when I have had some sleep.

IanSav:
I now understand the cause of the issue.  This is a result of a design flaw in the Enigma2 config engine.  Default values are not saved thus if the default changes the code can't tell that the previous default is not the same as the current default.

I am reworking the code to avoid this situation.

IanSav:
I think I have a solution to this issue as well as some further optimisation of the code.  Please try the code in [Pull Request #1557](https://github.com/openatv/enigma2/pull/1557).

PR #1564, opened 2020-03-16
noadsPiP - solution for filtering out ads

alec-cs:
Hi,
I have come up with the following solution for filtering out ads.

I have expanded and modified the existing PiP solution for Enigma2.

Solution steps (hopefully better understood):
1.) I see live TV broadcast on the TV screen.
2.) When I see an advertisement (ad) on the screen, I press a button on the remote control (PiP button or a programmed button).
3.) The active channel will be placed (moved) in the PiP window (with the ad)
4.) The following channel will be displayed on the main screen (Channel after the active channel).
5.) I see: Main screen (no ads) + PiP screen (with ads)
6.) When I see the ad end in the PiP window, I press a button on the remote control again (PiP button or programmed button).
7.) The contents of the PiP window will be displayed (moved) on the main screen and the PiP window will disappear.

This is how my solution would work.

The advertisement (ad) is viewed by sight, operation would be manual without automation!

My solution works fine on my system!
Usable and functional!


atvcaptain:
i check this in the next days

alec-cs:
Okay thank you.

If you have any questions please do not hesitate to contact me!

I have other plans and development ideas for expanding the PiP function. I would like to make this PiP operation more usable and functional.

If you have verified this change, I will write to you about my further ideas.
I've already made some of these on my system...

I'll wait for your feedback, your response.

Regards:
Alec

alec-cs:
Hi,

For ease of use, I also created a window to change the way PiP is used. If you press and hold the PiP button, the mode switching window will appear.
This makes it easier and faster to use different modes. It already works on my system. I've already done this.

I use the PiP function frequently in my system, so I want to make it more usable and better.
I say this based on my experience with using it.

Complementing the existing "standard" function, I am trying to improve the usage.
I created the "noadsPiP" function because of the filtering of the ads. I find this very useful and I use it a lot on a daily basis.

Another PiP feature I would like to implement is the "side-by-side" mode.
I would like to modify an existing similar mode as follows:
1.) After pressing the PiP (or programmed) button, the current channel will be displayed on the left side of the screen and the channel following the current channel on the right.
2.) The live broadcast on the left would be active (sound here), the live broadcast on the right would be inactive.
3.) You can change the Active status character by pressing the < and > buttons on the remote control, the live broadcast on the left or right should be active. Which live broadcast is active would be indicated by the word "Active" written below it.
    The left and right sides are not replaced, only the active status changes!
    All other operations affect only the active channel (channel change, EPG operations, ...).
4.) Pressing the EXIT button will only activate the live broadcast on the main screen, the inactive live broadcast will disappear from the screen.
    When you press the PiP (or programmed) button, the only inactive live broadcast will be on the main screen.

Obviously, this solution could be used if someone has the PiP feature.

This is how I thought about implementing and modifying the PiP function.
So there are 3 ways:
- standard -> Functions that are currently running
- noadspip -> Ads could be filtered out while ads go, another channel could be watched
- side-by-side -> If you want to watch multiple shows at once, you could use this mode the right way

The PiP mode switching window makes it easy to change the mode. No need to click multiple menu items in the main settings.

This is how I see these functions as usable and functional.
I think that is the way to implement them. It would be nice if it could work like this ...

I have already started to perform these functions, but only in the absence of information I am moving slowly. It takes time to look up and find out information.
It would be helpful if I had an answer to my questions in case of a problem.

If somebody were to help with development, they could be solved faster.

Hope what I wrote is understandable. If you have a question, ask me.
If I need to, I can take photos of my ideas.

I'm sorry I wrote so much.
  
I look forward to your feedback.

Regards:
Alec

atvcaptain:
now i have test and works as you say

can you rebase you pull request and we can merge it

thx for help make pip better

alec-cs:
Hi Captain,

Okay, first I will resolve the conflict so you can merge my development.
After that, I will add my "PiP mode selection window" upgrade to version 6.4.

A feature is not working properly in PiP. I couldn't fix the problem yet.
If I press "UP" or "DOWN" on the remote control, this function will disable PiP!
Why?  Please look at it!
I hope you understand what I mean. If you need me I'll send you a photo of that feature!

Other EPG and channel switching functions are working well, I have solved other problems.

Where can I ask my further development questions?
Where, in what forum can I ask my program code questions?

I need the information to develop my ideas faster.

Look forward to your reply.

Regards:
Alec

atvcaptain:
please send pull request only for noadsPiP - solution for filtering out ads

you see two wrong commit include too, send please new i close this

PR #1570, opened 2020-03-18
[Timezones.py] Multiple fixes

IanSav:
- Fix initialisation in "Classic" mode.
- Fix "posix" and "right" zoneinfo files not being properly suppressed.
- Fix the value of the "TZ" shell variable.
- Fix logging messages referring to "Directories".

### DIFF
--- a/lib/python/Components/Timezones.py
+++ b/lib/python/Components/Timezones.py
@@ -76,22 +76,27 @@ def InitTimeZones():
 		config.timezone.area.value = "Generic"
 	try:
 		tzLink = path.realpath("/etc/localtime")[20:]
-		tzSplit = tzLink.find("/")
-		if tzSplit == -1:
-			tzArea = "Generic"
-			tzVal = tzLink
-		else:
-			tzArea = tzLink[:tzSplit]
-			tzVal = tzLink[tzSplit + 1:]
 		msgs = []
-		if config.timezone.area.value != tzArea:
-			msgs.append("area '%s' != '%s'" % (tzArea, config.timezone.area.value))
-			# config.timezone.area.value = tzArea
-		if config.timezone.val.value != tzVal:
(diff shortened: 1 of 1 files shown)


Makumbo:
"Zeitzonengebiet" ist kein 'richtiges' Wort ;)
Es ist "doppelt gemoppelt"; eigentlich heißt es immer "Zeitzone", denn die "Zone" ist ja auch schon ein "Gebiet".
Und auch im englischen sollte es wohl eher "time zone" als "Timezone area" heißen.
"Your timezone in the area"? Eher "Your time zone", oder?
---
Shouldn't it be only "Time zone" instead of "Timezone area" because a 'zone' is also an 'area'?
And "Setup your timezone" instead of "Setup your timezone area"?
"Your timezone in the area"? Hmmm. Shouldn't it be "Your timezone" or "Your time zone"?
I'm confused.
Excuse my english.
Gruß - Greetings - Makumbo

IanSav:
How do you feel about these versions of the test:

Time zone area

    Select your time zone area or region.

Time zone

    Select the time zone within the area or region.

Makumbo:
I think "Select your time zone." does it. In **both** occasions.
In German "Wählen Sie Ihre Zeitzone."
So, maybe there is no need for two seperate texts.

Wikipedia says "A time zone is a region of the globe that ..."
So "Select your time zone area or region." or "Select the time zone within the area or region."
are 'repeated unnecessarily' (the term is I think).

It is not important where someone lives, in any case he is in a time zone...

Gruß - Makumbo

IanSav:
I will make the changes in the code.  The German will be up to the translators.  ;)

IanSav:
Done!

[Pull Request #1572](https://github.com/openatv/enigma2/pull/1572)

Makumbo:
Thanks!

PR #1601, opened 2020-04-07
Revert parameter corrections

IanSav:
- [skin.py] Revert parameter correction
  Revert the parameter parsing correction until all the places where single parameters are processed are fixed.

- [ConfigList.py] Revert parameter correction
  Revert the "separation" parameter parsing correction until all the places where single parameters are processed are also fixed.


IanSav:
Please don't merge this pull request until talking to me.

There is an inconsistency in the bug report that triggered this reversion.

PR #1608, opened 2020-04-13
Add Littlesat OpenPLi multiboot code plus additions for ATV

TwolDE2:
Please note that import from multiboot in SystemInfo MUST stay in position otherwise boot loop.
Apologies for some of the rubbish before todays commits!


atvcaptain:
i have update the patch and add to dev

PR #1620, opened 2020-04-27
Expanding the functionality of the PiP

alec-cs:
Hi,

I have expanded and modified the existing PiP solution for Enigma2.
I wanted it to be quick and easy to use.

Quick to use, either as a function or from a programmed button, or from the standard Blue button.
There is a warning window to set up a quick use of the Blue button.

The use of PiP and the PiP mode selection window can be assigned to a specific button on the remote control.
But it can be used just as quickly by the standard Blue button.

There are three modes to choose from to use PiP.
There are 3 ways:

  - standard -> Functions that are currently running
  - noadspip -> Ads could be filtered out while ads go, another channel could be watched
  - side-by-side -> If you want to watch multiple shows at once, you could use this mode the right way

The PiP mode switching window makes it easy to change the mode. No need to click multiple menu items in the main settings.

The main function of using PiP is Ad Filtering Mode. You can set how long it takes to be the default when it is in another mode. It can also be changed in the Menu.
You can also set the Color function buttons at the beginning or end of the list in the Extensions list. It can also be changed in the Menu.
This also helps with quick use.

You can also easily switch from Ad Filtering mode to Side-by-Side mode.

noadsPiP mode:
1.) I see live TV broadcast on the TV screen.
2.) When I see an advertisement (ad) on the screen, I press a button on the remote control (PiP button or a programmed button or Blue/Blue long).
3.) The active channel will be placed (moved) in the PiP window (with the ad)
4.) The following channel will be displayed on the main screen (Channel after the active channel).
5.) I see: Main screen (no ads) + PiP screen (with ads)
6.) When I see the ad end in the PiP window, I press a button on the remote control again (PiP button or a programmed button or Blue/Blue long).
7.) The contents of the PiP window will be displayed (moved) on the main screen and the PiP window will disappear.

The advertisement (ad) is viewed by sight, operation would be manual without automation!
I think this is a useful feature!

side-by-side mode:
1.) After pressing the PiP (or programmed) button, the current channel will be displayed on the left side of the screen and the channel following the current channel on the right.
2.) The live broadcast on the left would be active (sound here), the live broadcast on the right would be inactive.
3.) You can change the Active status character by pressing the LEFT button on the remote control, left and right live broadcasts can be swapped.
All operations affect only the active channel, left screen (channel change, EPG operations, ...).
4.) Press the RIGHT button to change the right screen with the Channel selection +/- buttons (Change inactive screen channel).
5.) Pressing the EXIT button will only activate the live broadcast on the main screen, the inactive live broadcast will disappear from the screen.
When you press the PiP (or programmed) button, the only inactive live broadcast will be on the main screen.

Other EPG and channel switching functions are working well, I have solved other problems.

standard mode:
- It works as usual, without change.

Multilingual usage is also resolved. I also created and added the .pot file.

I didn’t get much help to solve my development, but I managed to solve it somehow. I did it for a long time, but I feel like I’m done with it.
My development has been tested by testers on multiple set-top boxes and multiple SKINs. The tests were successful. Users found this development useful!

If you have any questions please do not hesitate to contact me!

Regards:
Alec

### DIFF
--- a/data/keymap.ntr
+++ b/data/keymap.ntr
@@ -1,4 +1,3 @@
-
 <keymap>
 	<map context="ListboxActions">
 		<device name="keyboard">
@@ -180,7 +179,7 @@
 
 	<map context="InfoBarQuickMenu">
 		<key id="KEY_BLUE" mapto="quickmenu" flags="b" />
- 	</map>
+	</map>
 
 	<map context="SatlistShortcutAction">
 		<key id="KEY_YELLOW" mapto="changetype" flags="b"/>
@@ -279,7 +278,6 @@
 		<key id="KEY_RIGHT" mapto="right" flags="mr" />
 	</map>
 
-
 	<map context="MenuActions">
 		<key id="KEY_MENU" mapto="menu" flags="mr" />
 		<key id="KEY_SPACE" mapto="menu" flags="mr" />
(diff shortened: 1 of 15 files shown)


atvcaptain:
ok i check it new

alec-cs:
Hi,

What do you think about my development?

Since then, I’ve added a little to the existing features. The "Channel list" (UP / DOWN) is already working. The EPG functions already work in the PiP window, there as well.

Any questions?

atvcaptain:
i am back, i check this today

PR #1654, opened 2020-05-16
remove not required string

thawtes:


atvcaptain:
i have clean up all po files

PR #1701, opened 2020-06-16
Update About.py

badsystem:


jbleyel:
Please do it linke so : _("%s GHz") % xxx

PR #1719, opened 2020-06-24
[KeyBindings.py] Unify codes across images

IanSav:
- Unify a coherent set of keyDescriptions, data and methods across images.
- Improve the documentation in the file.
- PEP8 clean up the file.

This change aligns the key mappings across images and will allow for improvements to the help system.

If there are any issues with the key mappings please ensure that the rcpositions.xml and index.html files are actually correct for your remote control.  Many of these files have incorrect button placements and mappings.


lib/python/Tools/KeyBindings.py
```diff
@@ -1,277 +1,379 @@
...
-		return [ ]
+		return []
 
 def getKeyDescription(key):
 	if rc_model.rcIsDefault():
 		idx = config.misc.rcused.value
 	else:
-		rctype = config.plugins.remotecontroltype.rctype.value
-		if rctype == 14:	# XP1000
+		# rcType = config.plugins.remotecontroltype.rctype.value
```
jbleyel, line 362: This will change the config value.

IanSav: Oh dear.  Sorry, I was unaware that the config variables also differed.

Do you want to correct this or should I?

jbleyel: Yes .. please let the variable as it is. For now.

IanSav: Okay, the error has been corrected in another commit to the pull request.

Thank you for bringing this error to my attention.  I will make the same correction to the other images.

PR #1761, opened 2020-07-17
Do not use list as variable in ConfigList

Taapat:
This fix error on ConfigList init.

For example in plugin YouTube init I get error:
    ConfigListScreen.__init__(self, searchList, session)
  File "/usr/lib/enigma2/python/Components/ConfigList.py", line 176, in __init__
    self["config"] = ConfigList(list, session = session)
  File "/usr/lib/enigma2/python/Components/ConfigList.py", line 21, in __init__
    self.list = _list
  File "/usr/lib/enigma2/python/Components/ConfigList.py", line 102, in setList
    assert len(x) < 2 or isinstance(x[1], ConfigElement), "entry in ConfigList " + str(x[1]) + " must be a ConfigElement"
AssertionError: entry in ConfigList None must be a ConfigElement

### DIFF
--- a/lib/python/Components/ConfigList.py
+++ b/lib/python/Components/ConfigList.py
@@ -18,7 +18,7 @@ class ConfigList(HTMLComponent, GUIComponent, object):
 		height, space = skin.parameters.get("ConfigListSlider", (17, 0))
 		self.l.setSlider(height, space)
 		self.timer = eTimer()
-		self.list = list
+		self.configList = list
 		self.onSelectionChanged = [ ]
 		self.current = None
 		self.session = session
@@ -111,7 +111,7 @@ class ConfigList(HTMLComponent, GUIComponent, object):
 
 	def isChanged(self):
 		is_changed = False
-		for x in self.list:
+		for x in self.configList:
 			is_changed |= x[1].isChanged()
 
 		return is_changed


jbleyel:
Do you have test this change?
"self.list" should not be a problem.

Taapat:
Yes, of course I test this change.
As you can see in the log "self.list" is a problem.
In the ConfigList class  is used list = property (getList, setList): https://github.com/openatv/enigma2/blob/6.5/lib/python/Components/ConfigList.py#L107
As you can see in the log "self.list = list" in init call this and therefore call the function setList, which on init does not need to run.
On Python 2 this does not happen, but on 3 there is a problem.

jbleyel:
I your log i see :
```
File "/usr/lib/enigma2/python/Components/ConfigList.py", line 21, in init
self.list = _list

```
But this should be :
```
File "/usr/lib/enigma2/python/Components/ConfigList.py", line 21, in init
self.list = list

```

jbleyel:
Do you mean this?
https://github.com/openatv/enigma2/blob/6.5/lib/python/Components/ConfigList.py#L107

Taapat:
Yes I pointed this to you.
Don't pay attention to the "_list", it's a log from the test series in which I tried to fix this by renaming the list.
But after that I concluded that the reason is exactly "self.list"

jbleyel:
Do we have the same issue here?

https://github.com/openatv/enigma2/blob/6.5/lib/python/Components/Sources/List.py#L27

Taapat:
There "self.list = list" is used on updateList.
I don't know where "self.list" is used after that, because this class doesn't use it.
I caught a previous problem on my plugin, so I was able to test the solution to fix it.
In this case, I think an additional test is needed.
But given the fact that this class does not use "self.list", I assume it will be used elsewhere, so there may be a problem changing its name.

atvcaptain:
i have revert the commit, openatv crash in the startup wizard, you issue youtube, i know i have fix other plugin with same issue, i must take a look

Taapat:
I alredy fix this in my plugin: https://github.com/Taapat/enigma2-plugin-youtube/commit/6163a6df4bce2f256b6e422a47137a2032f27936

atvcaptain:
ok thx for the info

PR #1798, opened 2020-08-06
Optimise image files

wedebe:
Optimise image files

This change reduces png, jpg and JPG image file sizes by passing
through ImageOptim lossless, "Extra" setting

"Saved 1.5 MB out of 5.9 MB. 37.8% per file on average (up to 96.6%)"


Schimmelreiter:
This was done on multiple E2 plugins in the past and it was reverted.
I dunno why, so let's see if someone else who was involved with it will answer.

wedebe:
> This was done on multiple E2 plugins in the past and it was reverted.
> I dunno why, so let's see if someone else who was involved with it will answer.

Worth a try. 1MB might not seem much to an end user but per image per box per download could save significantly on hosted bandwidth costs at the very least.

jbleyel:
ImageOptim is a very good tool to preserve space but pixmap cannot display all image formats. You need to test each changed image on your box first. I would prefer SVG instead of png. I don't know if e2 can use SVG for OSD rendering.

wedebe:
Hi @jbleyel , I haven't changed formats from what they were (png and jpg remain as-is, just with metadata fluff etc. removed).

Screen Shot 2020-08-06 at 22 54 27

Issue #1809, opened 2020-08-20
Strip po file references to filename only


wedebe:
Per
https://github.com/openatv/enigma2/commit/8c55d6916f7142c43ed1eaa8c19d246321866df5#commitcomment-41445871

po files aren't editable/diffable in the Github's web ui as they're now over the filesize threshold.

- ~Strip po file references to filename only~ #1819 
- ~File references should also be ; separated.~
  - let MSG tools as used in Makefile take care of this
- remove references to keymap and rcposition files as these aren't translatable

Could someone assign this to me, please?

PR #1810, opened 2020-08-20
Strip po file references to filename only and fix separators #1809

wedebe:
Strip po file references to filename only and fix separators #1809

In this change:
- Strip po file references to filename only (or pluginName/plugin.py)
- File references should also be ; separated.

Not part of this change:
- remove references to keymap and rcposition files as these aren't translatable


wedebe:
Noting that lt.po and nl.po didn't have file refs

wedebe:
Update to this PR in progress

PR #1845, opened 2020-10-04
6.4 dev

jeepcook:
Not sure using the good process for this pull request.
Concern 6.3, 6.4, 6.5 branches
File enigma2/lib/python/Screens/InfoBarGenerics.py

Wish: Add channel name in groupedservices screen.  I think the modification has to be made at line 207:
current_show_name = title + " " + str(starttime) + "-" + str(endtime)

So currently we have show name and start/end time of the show, but no channel name. But it will be more readable if the channel name (service) was displayed somewhere, ideas: at the beginning of the subservice line or at the end with ().

For example, currently for the group HD Plus set in /usr/share/enigma2/groupedservices we have a lot of different channels (not declination of just one channel), and without displaying channel name the choice is not easy to do, it's not realy readable.
 
Hope my request is understanding.

Thanks

### DIFF
--- a/data/keymap.xml
+++ b/data/keymap.xml
@@ -120,8 +120,8 @@
 		<!-- Also uses NavigationActions -->
 		<!-- Also uses NumberActions -->
 		<!-- Also uses TextEditActions -->
-		<key id="KEY_RED" mapto="cancel" flags="m" />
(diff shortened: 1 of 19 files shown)


nickersk:
please, rework your pull request to match its description. thank you.

PR #1853, opened 2020-10-20
Update uk.po

badsystem:


nickersk:
line 942 syntax error: keyword "a" unknown

PR #1866, opened 2020-11-24
Tryfix spinner after recordings

4l3x2k:
Thanks to @SimonCapewell for OpenViX

https://github.com/OpenViX/enigma2/commit/3172dd4b978dcd692cb8291b6c3e95c718d2b661
https://github.com/OpenViX/enigma2/commit/d93ddcd81f25972e6b4933278d6ca5f52806286c

https://www.opena.tv/openatv-6-4-rueckmeldungen/50414-taeglich-abstuerze-meistens-nach-aufnahmen.html

### DIFF
--- a/lib/dvb/filepush.cpp
+++ b/lib/dvb/filepush.cpp
@@ -46,6 +46,12 @@ static void signal_handler(int x)
 
 static void ignore_but_report_signals()
 {
+	/* we must set a signal mask for the thread otherwise signals don't have any effect */
+	sigset_t sigmask;
+	sigemptyset(&sigmask);
+	sigaddset(&sigmask, SIGUSR1);
+	pthread_sigmask(SIG_UNBLOCK, &sigmask, NULL);
+	
 	/* we set the signal to not restart syscalls, so we can detect our signal. */
 	struct sigaction act;
 	act.sa_handler = signal_handler; // no, SIG_IGN doesn't do it. we want to receive the -EINTR
@@ -537,16 +543,11 @@ int eFilePushThreadRecorder::read_dmx(int fd, void *m_buffer, int size)
 
 void eFilePushThreadRecorder::thread()
 {
+	ignore_but_report_signals();
+	
 	setIoPrio(IOPRIO_CLASS_RT, 7);
 
 	eDebug("[eFilePushThreadRecorder] THREAD START");
-
-	/* we set the signal to not restart syscalls, so we can detect our signal. */
-	struct sigaction act;
-	memset(&act, 0, sizeof(act));
-	act.sa_handler = signal_handler; // no, SIG_IGN doesn't do it. we want to receive the -EINTR
-	act.sa_flags = 0;
-	sigaction(SIGUSR1, &act, 0);
 
 	hasStarted();
 	if (m_protocol == _PROTO_RTSP_TCP)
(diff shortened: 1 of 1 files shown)


atvcaptain:
have you test it and works ? vix commits show many more changes

SimonCapewell:
Yes, it looks like you're missing the extra exit conditions in eFilePushThreadRecorder::thread method. That's the more important change because, with just the SIGUSR1 fix in place, `poll` could consume the signal and cause `read` to block indefinitely.

4l3x2k:
@atvcaptain 
No. Not tested. Build system is doing other things and I do not have any target device that is affected by this problem.

@SimonCapewell 
I actually do not see the "extra exit" condition. What do you mean? This enigma2 do not have that BC encoder poll code at all. The only sys call that should be interrupted is read. And the m_stop is evaluated after that.

That is why I only added the signal mask for that particular thread. I only think that the order of signal mask and handler registration is wrong. It should be handler registration and then enable the signal.

SimonCapewell:
These are the lines you need https://github.com/OpenViX/enigma2/blob/61f1bca6121649cc0ee7c15a93e92ebbf8f206c2/lib/dvb/filepush.cpp#L351-L360

_poll_ is a syscall, so it can consume the signal, meaning that there's a risk of the subsequent call to _read_ blocking indefinitely if the read buffer is empty. There's only a tiny chance of it happening, but with the extra checks for _m_stop_, there's no chance of it happening.

4l3x2k:
@SimonCapewell 
Ok. I missed that. I thought read should be interrupted by signal. Seems I'am a bit confused by the manpage of read.
Thank you for explanation.

SimonCapewell:
Yes, both can be interrupted by a signal. For testing purposes you can replace `read` or `poll` with `pause` to see the signal in action

nickersk:
seems ok to me.

4l3x2k:
Someone really needs to test with an affected box as I do not want to be responsible for a headache of all the oatv users :).

I'll ask in the board whether someone has this opportunity.

4l3x2k:
@atvcaptain 
At least for the last commit of this PR there is a test: https://www.opena.tv/openatv-6-4-rueckmeldungen/50414-taeglich-abstuerze-meistens-nach-aufnahmen-post451094.html#post451094
I could add poll.h and remove first commit.

4l3x2k:
I'am closing this PR because:

- there are more / other changes required
- missing tests
- only a few mipsel boxes are affected
- these boxes are only affected if recording is used

The risk of changing 6.4 stable and drive users crazy that do not own such box nor do recordings is too high. Those who own such box should use 6.3 or wait for 6.5.

SimonCapewell:
I've just realised you were pointing out that you didn't have the poll pulled in from PLi. Sorry! I think I hadn't realised this was ATV when I commented 🤦 For my ET8500 on Vix, `poll` was the trigger for the recording spinner problems, which led to discoving that signals weren't exiting the read either.

4l3x2k:
@SimonCapewell 
I see.
That's why I said "This enigma2 do not have that BC encoder poll code at all."
Poll has a timeout set by the way. Unlikely it will block. But it is interrupted and the following read would block. That makes sense.
I did this first commit of this PR because I want to be sure the signal is enabled for this thread. We're missing THREAD STOP after signal is sent.
Hopefully the signal is delivered. I think that normal kill is not thread aware and delivers to main thread - the process - which might ignore the signal. Something like sigqueue is needed.
What's ever open is the "freaky" async writing ... I didn't get there.

Thank you for joining the conversation!

PR #1874, opened 2020-12-13
filepush - stop ignoring SIGUSR1 leading to endless read()

nickersk:
thx @SimonCapewell and @eriksl

- https://github.com/OpenViX/enigma2/commit/3172dd4b978dcd692cb8291b6c3e95c718d2b661#
- https://github.com/OpenPLi/enigma2/commit/ee43886a259cda41f933a8f3f490472cef84453f

pls dont merge untill tested on various machines
https://www.opena.tv/openatv-6-4-rueckmeldungen/50414-taeglich-abstuerze-meistens-nach-aufnahmen-54.html

### DIFF
--- a/lib/dvb/filepush.cpp
+++ b/lib/dvb/filepush.cpp
@@ -3,6 +3,8 @@
 #include <errno.h>
 #include <fcntl.h>
 #include <sys/ioctl.h>
+#include <poll.h>
+#include <signal.h>
 
 #if defined(__sh__) // this allows filesystem tasks to be prioritised
 #include <sys/vfs.h>
@@ -16,6 +18,11 @@
 //#define SHOW_WRITE_TIME
 
 DEFINE_REF(eFilePushThread);
+static void global_signal_SIGUSR1_handler(int x)
+{
+	eWarning("Signal SIGUSR1 caught\n");
+}
+
 eFilePushThread::eFilePushThread(int io_prio_class, int io_prio_level, int blocksize, size_t buffersize)
 	: prio_class(io_prio_class),
 	  prio(io_prio_level),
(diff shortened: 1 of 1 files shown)


4l3x2k:
I'am not really happy with this because I do not see how read would be interrupted. This is meant to the first call to read before (p)poll.

The only non-blocking read call is read_dmx which is for streaming because the file descriptor is flagged with O_NONBLOCK for that protocol.

The (p)poll call wouldn't even need to be interrupted because it has a timeout.

The read call needs to be guarded by the (p)poll call like it is in openvix (m_stop is evaluated) or openpli (having if then else around the result of (p)poll). Only then it makes sense to me.


One real abstract question to me is whether the read of demux would ever block?! Maybe only if no filter is set. The demux could deliver garbage but it would deliver something. So read would never block!
One other point would be that in my knowledge of C - not C++ - the action and sigset are going to be lost as they are only declared and defined in function blocks and are local to these. There is no guarantee they live as long as the thread does. So these two things would need to be in class declaration.

Sorry to be so ... pedantic. But this is the second approach / try to get these spinner after recordings fixed. Great thanks to @nickersk!

SimonCapewell:
Yes, testing shows this to still be buggy. We now have a race condition where `ppoll` returns `1`, but `pfd.revents != POLLIN` leading to the `read` blocking indefinitely if `m_source_fd` is no longer being fed with data. Given that the demuxer is shut down _before_ the thread is stopped, this leads to a window where there's no data to read. All it then takes is a soft error condition from `ppoll` for the code to fall through to the read. Plus, now that signals are blocked everywhere except the `ppoll`, the thread stop signal won't break out of a blocked read!

4l3x2k:
@nickersk @SimonCapewell
@eriksl is working it out. Another fix for the fix :-D.

I know this is really hard stuff here.
Just because a blocking read in a less meaningful thread that could be termed and then killed without any respect.

PR #1906, opened 2021-04-10
py3 compatible

persianpros:
@jbleyel 

We use this in OV as py2/py3 compatible, is it ok with 6.5 which is for py3 only?

### DIFF
--- a/lib/python/Components/ActionMap.py
+++ b/lib/python/Components/ActionMap.py
@@ -99,7 +99,7 @@ class HelpableActionMap(ActionMap):
 					print("[HelpActionMap] removed duplicity: %s %s" % (context[1], record))
 					return True
 			return False
-		if not type(contexts) is list:
+		if not isinstance(contexts, list):
 			contexts = [contexts]
 		actions = actions or {}
 		self.description = description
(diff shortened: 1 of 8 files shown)


jbleyel:
6.5 ist NOT py3 only!!

jbleyel:
This needs to be testet first for py2 and py3

persianpros:
Then it should be fine as we're using for py2 for now.

jbleyel:
has_key and isinstance should be no difference.
I'm not sure about sorted list result for py3.

jbleyel:
Hi @atvcaptain , please check.

persianpros:
@jbleyel 

Should I send the PEP8 for 6.4 branch? As you merge from 6.4 to 6.5 I think it's needed but your call.

jbleyel:
Why not. I have done this for 6.5 already.

jbleyel:
PY3 Docu:

You can also use the list.sort() method. It modifies the list in-place (and returns None to avoid confusion). Usually it’s less convenient than sorted() - but if you don’t need the original list, it’s slightly more efficient.

jbleyel:
Please make a new pull but do only the has_key changes.

PR #1909, opened 2021-04-11
Use futurize to fix has_key

persianpros:


jbleyel:
Hi @persianpros, nice work but I think we will not apply these changes on py2 only branches.

persianpros:
First I sent it for your main branch then you can cherry-pick it to 6.5 or I can send another for that branch :+1:

jbleyel:
I will not make merge because of too many and untested changes. You may have no effects but it's untested. It's to dangerous.

persianpros:
You could test it and merge it later, up to you :+1: 

We use it for py2 and it's ok, I'll send another PR for 6.5 soon.

jbleyel:
Hi @persianpros , i will do the change on 6.5. Thanks

PR #1919, opened 2021-04-18
Use futurize to fix has_key

persianpros:
Safe for py2 and py3, similar to https://github.com/OpenViX/enigma2/pull/673

It only changes has_key to "in" no other changes.

After this PR all PLi, ViX, Vision and ATV would be similar and without has_key.

### DIFF
--- a/lib/python/Components/Network.py
+++ b/lib/python/Components/Network.py
@@ -107,14 +107,14 @@ class Network:
 		fp.write("auto lo\n")
 		fp.write("iface lo inet loopback\n\n")
 		for ifacename, iface in self.ifaces.items():
-			if iface.has_key('dns-nameservers') and iface['dns-nameservers']:
+			if 'dns-nameservers' in iface and iface['dns-nameservers']:
 				dns = []
 				for s in iface['dns-nameservers'].split()[1:]:
 					dns.append((self.convertIP(s)))
 				if dns:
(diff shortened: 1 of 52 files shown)


lib/python/Components/Network.py
```diff
@@ -557,7 +557,7 @@ def deactivateInterfaceFinished(self, extra_args):
 		(ifaces, callback) = extra_args
 
 		def checkCommandResult(iface):
-			if self.deactivateInterfaceConsole and self.deactivateInterfaceConsole.appResults.has_key("ifdown " + iface):
+			if self.deactivateInterfaceConsole and "ifdown " + iface in self.deactivateInterfaceConsole.appResults:
```
jbleyel, line 560: this should be `
if self.deactivateInterfaceConsole and ("ifdown " + iface) in self.deactivateInterfaceConsole.appResults:`

or what you think?

persianpros: You mean having parentheses around the check to make sure it checks both?

I agree with you but is the result any different

jbleyel: yes .. i'ts the same result. I have test this.

persianpros: So we're good :+1: 

I need to send this for other distros, this is the first time e2 scene is getting same style.

PR #1926, opened 2021-04-26
Introduce getPyExt to let us know when we should read pyo and when py (python 3)

persianpros:
This is the combine of https://github.com/OpenVisionE2/enigma2-openvision/commit/6dd8338124db149f2e2268c8dc2287904ef065ee and https://github.com/OpenVisionE2/enigma2-openvision/commit/83948f1b55a1f9553729b8ca0d4967ba25d85819 for OpenATV

Simple example of how to use "getPyExt": https://github.com/OpenVisionE2/Internet-Speedtest/commit/03e5ec2fcb796061049cb5d9bb6b6f3b6b5c17a5#diff-313cb6c49ded3951ffdf161f19c69faf48996d6bbb85f41d73bfb53809b81cf1

Unfortunately ATV is full of hard-coded "/usr/lib/enigma2/python/Plugins/" which should be replaced by 'resolveFilename(SCOPE_PLUGINS, "' so the next good cleanup for ATV could be this.

With getPyExt we no longer need to use "or conditions" or import six for a simple detection.

Even a simple check like:

if getPyExt() == "pyo":
pyversion = 2
else:
pyversion = 3

could help you instead of

if six.PY2 or if six.PY3

For this to work we need OE support as well: https://github.com/oe-alliance/oe-alliance-core/pull/672

### DIFF
--- a/lib/python/Components/Converter/TrackInfo.py
+++ b/lib/python/Components/Converter/TrackInfo.py
@@ -3,6 +3,7 @@ from Components.Element import cached
 from Poll import Poll
 from Tools.ISO639 import LanguageCodes
 from Tools.Directories import fileExists
+from enigma import getPyExt
 
 
 class TrackInfo(Poll, Converter, object):
@@ -95,7 +96,7 @@ class TrackInfo(Poll, Converter, object):
 					kodi = KodiVideoPlayer.instance
 				except:
 					kodi = None
-				if kodi and fileExists("/usr/lib/enigma2/python/Plugins/Extensions/SubsSupport/plugin.pyo"):
+				if kodi and fileExists("/usr/lib/enigma2/python/Plugins/Extensions/SubsSupport/plugin.", getPyExt()):
 					if kodi.embeddedEnabled:
 						selectedSubtitle = kodi.selected_subtitle
 						enabled = kodi.subtitle_window.shown
--- a/lib/python/Components/Converter/TrackInfo.py
+++ b/lib/python/Components/Converter/TrackInfo.py
@@ -3,7 +3,6 @@ from Components.Element import cached
 from Poll import Poll
 from Tools.ISO639 import LanguageCodes
 from Tools.Directories import fileExists
-from enigma import getPyExt
 
 
 class TrackInfo(Poll, Converter, object):
@@ -96,7 +95,7 @@ class TrackInfo(Poll, Converter, object):
 					kodi = KodiVideoPlayer.instance
 				except:
 					kodi = None
-				if kodi and fileExists("/usr/lib/enigma2/python/Plugins/Extensions/SubsSupport/plugin.", getPyExt()):
+				if kodi and fileExists("/usr/lib/enigma2/python/Plugins/Extensions/SubsSupport/plugin.pyo"):
 					if kodi.embeddedEnabled:
 						selectedSubtitle = kodi.selected_subtitle
 						enabled = kodi.subtitle_window.shown
--- a/lib/python/Components/PluginComponent.py
+++ b/lib/python/Components/PluginComponent.py
@@ -56,7 +56,7 @@ class PluginComponent:
 						except Exception, exc:
 							print "Plugin ", c + "/" + pluginname, "failed to load:", exc
 							# supress errors due to missing plugin.py* files (badly removed plugin)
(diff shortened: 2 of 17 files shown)


persianpros:
I hope all OE-A accept this so we could ask PLi to follow it also.

Although they don't have a plan for python 3 but at least it helps to avoid using hard-coded variables.

lib/python/Components/PluginComponent.py
```diff
@@ -56,7 +56,7 @@ def readPluginList(self, directory):
 						except Exception, exc:
 							print "Plugin ", c + "/" + pluginname, "failed to load:", exc
 							# supress errors due to missing plugin.py* files (badly removed plugin)
-							for fn in ('plugin.py', 'plugin.pyc', 'plugin.pyo'):
+							for fn in ('plugin.py', 'plugin.pyo'):
```
jbleyel, line 59: This should not be changed.

lib/python/Components/Converter/TrackInfo.py
```diff
@@ -95,7 +96,7 @@ def getText(self):
 					kodi = KodiVideoPlayer.instance
 				except:
 					kodi = None
-				if kodi and fileExists("/usr/lib/enigma2/python/Plugins/Extensions/SubsSupport/plugin.pyo"):
+				if kodi and fileExists("/usr/lib/enigma2/python/Plugins/Extensions/SubsSupport/plugin.", getPyExt()):
```
jbleyel, line 99: Are you sure about "string" , getPyExt().
This looks like a second parameter for fileExists.

lib/python/Plugins/Extensions/Infopanel/QuickMenu.py
```diff
@@ -61,7 +61,7 @@
 else:
 	AUDIOSYNC = False
 
-if path.exists("/usr/lib/enigma2/python/Plugins/SystemPlugins/VideoEnhancement/plugin.pyo"):
+if path.exists("/usr/lib/enigma2/python/Plugins/SystemPlugins/VideoEnhancement/plugin.", getPyExt()):
```
jbleyel, line 64: Are you sure about "string" , getPyExt().
This looks like a second parameter for path.exists.

jbleyel (commented):
Please do not change the py code with this commit.

lib/python/Components/PluginComponent.py
persianpros: As far as I know pyc is ancient and we removed it from PLi and Vision but I can keep it for ATV if you want.

lib/python/Components/Converter/TrackInfo.py
persianpros: No I'm not sure as I can't test it because we don't have it in OV.

We don't use hard-coded path and I need your help testing it and give me the correct syntax.

Once I get the what's right I can fix them all :+1:

jbleyel: Please let all the py files as they are.

jbleyel:
```
>>> os.path.exists("/tmp", 's')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: exists() takes exactly 1 argument (2 given)
>>>
```

persianpros:
@jbleyel 

Python changes reverted, I got what's the problem.

We use fileExists(resolveFilename and that's why we don't see those errors as we pass the result of resolveFilename to fileExists.

But now you know where are those needed pyo lines ;)

Please squash merge this only.

jbleyel:
@betacentauri , @atvcaptain .. please review.

persianpros:
BTW as now you know where are some of hard-coded things it's a good time to change them :+1:

atvcaptain:
I don't see any advantages if you use something like that , especially with python3 we will certainly switch to pyc in some cases

atvcaptain:
I don't see any advantages if you use something like that , especially with python3 we will certainly switch to pyc in some cases

jbleyel:
I like this -> 
https://github.com/openatv/enigma2/commit/384032df3d9c6ef40d6ac0fc2138d200c5c4d4dd

persianpros:
https://github.com/openatv/enigma2/commit/384032df3d9c6ef40d6ac0fc2138d200c5c4d4dd is better :+1: 

Please use this in all distros.

PR #1971, opened 2021-07-09
Sync of OV and ATV

jbleyel:
This pull will sync some files of OV and ATV and should not change anything.
* add m_long_key_pressed ivar
* add function parsePts_t
* add eTraceNoNewLineStart and eTraceNoNewLine
* use template class for ePtrList erase
* add getNoWrap to eLabel
* prevent eWidget crashes if root is null
* changes some log outputs


betacentauri:
Looks ok for me.

nickersk:
for me ok too

PR #1976, opened 2021-07-12
Text tweaks - `Couldn't record...`

wedebe:


jbleyel:
Nice work, but.
Could you please make one big commit and not too many small.

PR #1984, opened 2021-07-23
update config.py

jbleyel:
* use new ACTIONKEY_ constants
* cosmetics
* refactor ConfigSequence

Thanks to @IanSav

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -401,7 +401,7 @@ class ConfigBoolean(ConfigElement):
 		elif key == ACTIONKEY_LAST:
 			self.value = True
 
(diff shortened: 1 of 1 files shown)


atvcaptain:
small typo if self.censor_char == "" or not self.hidden:: -> ::

atvcaptain:
next typo 
  File "/usr/lib/enigma2/python/Components/config.py", line 405, in fromstring
    return str(value).lower() in self.trueValues
NameError: name 'value' is not defined

cange def fromstring(self, val): to def fromstring(self, value):

jbleyel:
Yep .. i'm on.

PR #2017, opened 2021-08-21
Update fi.po

tomzza:
Adding missing, and change better match.


atvcaptain:
please use https://hosted.weblate.org/engage/openatv/ to add new Translation thx

PR #2030, opened 2021-08-22
Revert "[ConfigDirectory] add missing value properties"

4l3x2k:
This reverts commit 1a2e392dfcaef0118889f540e6d2157b4e1a1f38.

Breaks FileCommander and EMC


jbleyel:
We need to fix FileCommander and EMC!!!

PR #2119, opened 2021-11-16
Control a motor with an external command

olivluca:
This modification adds support for moving the antenna using
an external command to be used with
https://github.com/olivluca/actuino.
Once the receiver has been configured as usals, if there is an
executable file /home/root/rotor_X (where X is the number of the frontend)
it will be used to start moving the rotor (it will be called with
two parameters: the orbital position and the stored position).
This command should terminate quickly (i.e. just start the movement,
not wait for the movement to complete).
It will also check for the existence of the file /tmp/rotor_moving_X
to see it the motor is still moving.


olivluca:
Thank you.

PR #2149, opened 2021-12-25
Add Inverto ISKY-UST110-CUO4O-16P

AbuBaniaz:
```http://www.inverto.tv/multiswitch/307/unicable-ii-cascadable-switch-with-terrestrial-input-and-4x-sky-dscr-satcr-legacy-terrestrial-outputs```


AbuBaniaz:
This only works in format=diseqc mode.

PR #2150, opened 2021-12-26
Update nl.po

Snoete:


jbleyel:
Hello @Snoete , you can use weblate for 7.0

PR #2206, opened 2022-02-24
cleanup

geeeeeeeeeee:
Who need a glasses? I see twice.
The beginning: https://github.com/openatv/enigma2/commit/33b514d049ce8f2e576c613f704c02559b32a905
Please check my code first or try it self but i found this... and this should be the same.


geeeeeeeeeee:
Sorry 
def getUpdates(self, callback = None):   -->  callback=None
Check formatting too please.

PR #2207, opened 2022-02-24
cleanup v2

geeeeeeeeeee:
Who need a glasses? I see twice.
The beginning: https://github.com/openatv/enigma2/commit/33b514d049ce8f2e576c613f704c02559b32a905
Please check my code first or try it self but i found this... and this should be the same.


geeeeeeeeeee:
My code wasn't good but i cant pull nothing and i know u can do it better and i cant fully understand it at the moment.
Good idea to report? On this way?
Thank you.

PR #2223, opened 2022-03-14
[Setup.py] Correct setup element list

IanSav:
This change forces the self["config"] list to be cleared if there are no eligible items available to be displayed.


IanSav:
This code is from a future version of Setup.py and does not work in this version.

PR #2261, opened 2022-04-08
Fix sending datetime to CAM

catalinii:
This code seems very similar to https://github.com/atvcaptain/dvb-apps/blob/master/lib/libdvben50221/en50221_app_datetime.c#L100 but https://github.com/atvcaptain/dvb-apps/blob/9b74778ab215bbe8901cac7ff176d039c725a23c/lib/libdvben50221/en50221_session.c#L280 does not prepend the length of the packet to the data sent to the CAM.
So for enigma2 this ends up as:
90 02 00 04 9f 84 41 06 05 e9 1d 20 11 52
instead of:
90 02 00 04 9f 84 41 05 e9 1d 20 11 52

### DIFF
--- a/lib/dvb_ci/dvbci_datetimemgr.cpp
+++ b/lib/dvb_ci/dvbci_datetimemgr.cpp
@@ -63,14 +63,14 @@ void eDVBCIDateTimeSession::sendDateTime()
 	tv %= 60;
 	uint8_t ss = tv;
 
-	msg[0] = 5; // not using offset
-	msg[1] = (mjd >> 8) & 0xff;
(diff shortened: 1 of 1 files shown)


Papi2000-new:
Is it possible, that the same topic also exists in oATV7.0 or oATV6.4, and hinders the module working correctly?

jbleyel:
That's ATV 7.0 and 7.1 .. we do not have an extra branch for 7.1 yet.
6.4 is different and needs to check separately.

catalinii:
I think this needs to be backported to all supported branches as this code exists everywhere.

PR #2271, opened 2022-04-20
[TimerEdit]

jbleyel:
* refactor imports
* use StaticText for color buttons in TimerEditList and TimerSanityConflict


lib/python/Screens/TimerEdit.py
```diff
@@ -38,80 +39,82 @@ class TimerEditList(Screen):
 
 	def __init__(self, session):
 		Screen.__init__(self, session)
-		Screen.setTitle(self, _("Timer List"))
-
+		self.setTitle(_("Timer List"))
```
IanSav, line 42: Is this the correct title...

lib/python/Screens/TimerEdit.py
```diff
@@ -38,80 +39,82 @@ class TimerEditList(Screen):
...
+			"green": self.addCurrentTimer,
+			"log": self.showLog,
+			# "top": self.moveTop,
+			"left": self.pageUp,
+			"up": self.moveUp,
+			"down": self.moveDown,
+			"right": self.pageDown,
+			# "bottom": self.moveBottom
+		}, prio=-1)
 		self.setTitle(_("Timer overview"))
```
IanSav, line 69: Or is this the correct title?  If so, it should be "Timer Overview" and the earlier setTitle() should be removed.

IanSav (approved):
Thanks.

PR #2283, opened 2022-05-03
Add support for enabling and disabling CI devices

catalinii:
- By default every device is enabled
- Allows the user to disable the CI device from the menu (to be used by other software)
- At startup a disabled device will not initialize but will appear in the menu as disabled and the user can choose to enable it

This functionality triggers also a Reset. @atvcaptain do you think we should remove the Reset functionality or leave it as it is.

### DIFF
--- a/lib/python/Screens/Ci.py
+++ b/lib/python/Screens/Ci.py
@@ -33,6 +33,8 @@ def setCIBitrate(configElement):
 	else:
 		eDVBCI_UI.getInstance().setClockRate(configElement.slotid, eDVBCI_UI.rateHigh)
 
+def setCIEnabled(configElement):
+    eDVBCI_UI.getInstance().setEnabled(configElement.slotid, configElement.value)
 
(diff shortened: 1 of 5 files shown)


jbleyel (commented):
Why not Boolean as enabled parameter for the function?

catalinii:
No reason at all... I am more used to C and ints as booleans. I can refactor if needed

lib/python/Screens/Ci.py
```diff
@@ -19,6 +19,7 @@
 from Tools.BoundFunction import boundFunction
 from boxbranding import getBrandOEM
 import time
+import traceback
```
jbleyel, line 22: do we need this

jbleyel:
> No reason at all... I am more used to C and ints as booleans. I can refactor if needed

I'ts better coding style imho.

catalinii:
@jbleyel refactored int -> bool

PR #2292, opened 2022-05-07
[VolumeControl] `Temporary mute` functionality

wedebe:
[VolumeControl] `Temporary mute` functionality

This change introduces a short-term (30s) mute action on `Mute` button hold
(suitable for annoying ads etc.)

This change does not:
- have user-configurable time


IanSav:
If you are going to press a button to mute the volume then why not press another button to skip forwards 30 seconds and get rid of the commercial completely?  (You *can* configure the skips to eliminate a range of commercial break durations.)

I don not believe change of MUTE functionality is appropriate.

wedebe:
> ... why not press another button to skip forwards 30 seconds and get rid of the commercial completely?

Because I don't (yet) have access to each and every linear/live broadcasters' playout centres to skip through the ads. I'm sure other viewers would be confused as to why I've clumsily skipped part-way through an ad break and reduced their time to make a cup of tea or go for a pee 😬

PR #2302, opened 2022-05-09
Remove DualMode because it's now part of Multiboot

jbleyel:


IanSav (approved):
I can't test any of this but the changes look okay to me.

PR #2308, opened 2022-05-14
[FlashManager] - fix Crash on FlashImage

BliNeR-KeY:


lib/python/Screens/FlashManager.py
```diff
@@ -282,7 +282,9 @@ def __init__(self, session, imagename, source):
 		self["progress"] = ProgressBar()
 		self["progress"].setRange((0, 100))
 		self["progress"].setValue(0)
-		self.callLater(self.confirmation)
+		self.delay = eTimer()
```
jbleyel, line 285: This is the same as self.callLater

```
	def callLater(self, function):
		self.__callLaterTimer = eTimer()
		self.__callLaterTimer.callback.append(function)
		self.__callLaterTimer.start(0, True)

```

PR #2375, opened 2022-06-16
[config.py] force utf-8 on write of settings

jbleyel:


IanSav (approved):

Huevos:
_encoding is the name of the encoding used to decode or encode the file. This should only be used in text mode. The default encoding is platform dependent (whatever [locale.getpreferredencoding()](https://docs.python.org/3/library/locale.html#locale.getpreferredencoding) returns), but any [text encoding](https://docs.python.org/3/glossary.html#term-text-encoding) supported by Python can be used._

[https://docs.python.org/3/library/functions.html#open](url)

```
root@vuultimo4k:~# python
Python 3.10.4 (main, Mar 23 2022, 20:25:24) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
>>> import locale
>>> locale.getpreferredencoding()
'UTF-8'
>>>
```
Which means with `open("%s.writing" % filename, "w") as fd:` and `with open("%s.writing" % filename, "w", encoding="UTF-8") as fd:` are identical.

jbleyel:
>>> locale.getpreferredencoding()
'UTF-8'

Huevos:
Ok, so you have proved your own argument then.

jbleyel:
> Ok, so you have proved your own argument then.

No really .. 

```
locale.getpreferredencoding()
'UTF-8'

```

Huevos:
Yes, I understand. Python is not complying with what it says in the manual.

jbleyel:
The force to utf-8 should be no issue because all python 3 boxes should have uff-8 as default.

nautilus7:
> The force to utf-8 should be no issue because all python 3 boxes should have utf-8 as default.

That is correct. Setting the encoding explicitly should not harm.

The crash in your docker image should be caused if enigma2 is started without having set a utf8 locale.
How do you start enigma2? Do you use init 3 or simply with enigma2 command? Make sure LC_CTYPE is set to a utf8 locale in enigma2 start script.

OpenPLi had a similar issue. Read: https://github.com/OpenPLi/enigma2/commit/198d8e53ce2cd158a6a52adca514266c6e06d11f

PR #2395, opened 2022-06-27
[ImageWizard]

jbleyel:
* refactor and cleanups


IanSav (changes requested):
Improvements and changes have been provided via the Skype channel.

IanSav (changes requested):
I would still like to see the hardware conditional to be abstracted from dat in SystemInformation.py.  This would probably be via a BoxInfor variable as I suggested.

PR #2401, opened 2022-06-29
[MultiBoot.py] Use Conversions.py date formatting

IanSav:
This change uses the "formatDate()" date formatting method from "Tools/Conversions.py" to format the slot date.

NOTE: This code needs the import to be local to its method to avoid creating a boot loop.  This is a cosmetic change and does not fix any issues or change any functionality.  The change is totally optional.  If this import location requirement is seen to be undesirable then this pull request should not be merged.


IanSav:
The date formatting improvement has not been deemed significant enough to use.

PR #2406, opened 2022-07-01
use driversdate from enigma.info

jbleyel:
[Information]
* use driversdate from enigma.info
* use formatDate for imagebuild

### DIFF
--- a/lib/python/Screens/Information.py
+++ b/lib/python/Screens/Information.py
@@ -749,7 +749,7 @@ class DistributionInformation(InformationBase):
 		if self.extraSpacing:
 			info.append("")
 		info.append(formatLine("P1", _("Distribution"), BoxInfo.getItem("displaydistro")))
-		info.append(formatLine("P1", _("Distribution build"), BoxInfo.getItem("imagebuild")))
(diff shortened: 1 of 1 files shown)


lib/python/Screens/Information.py
```diff
@@ -1411,7 +1411,7 @@ def findPackageRevision(package, packageList):
 		info.append(formatLine("S", _("Driver and kernel information")))
 		if self.extraSpacing:
 			info.append("")
-		info.append(formatLine("P1", _("Drivers version"), formatDate(about.getDriverInstalledDate())))
+		info.append(formatLine("P1", _("Drivers version"), formatDate(BoxInfo.getItem("driversdate"))))
```
IanSav, line 1414: Does this mean that "about.getDriverInstalledDate()" can also now be deleted?

IanSav (approved):
It would be good if even more of the "About.py" module can be deleted and/or moved to SystemInfor.py.

lib/python/Screens/Information.py
BliNeR-KeY: driversdate on enigma-kernel module fixed by @atvcaptain 20 days ago. working as well
https://github.com/oe-alliance/oe-alliance-core/commit/f5e9e7b5bb0e0ad5466fcb347dd20a97548154e2

PR #2408, opened 2022-07-01
[Menu]

jbleyel:
* allow to enable/disable menu description


IanSav:
This can, and does, create issues for some skins.  On MetrixHD this is just some text overlayed on existing screen elements so making the text go away has no significance.  This may not be the case for all skins.  For example, on OverlayHD the menu list is shortened to make space for the description.  When the description is turned off the screen looks broken.

IanSav:
This can, and does, create issues for some skins. On MetrixHD this is just some text overlayed on existing screen elements so making the text go away has no significance. This may not be the case for all skins. For example, on OverlayHD the menu list is shortened to make space for the description. When the description is turned off the screen looks broken.

PR #2416, opened 2022-07-03
Picture player

jbleyel:


IanSav (approved):
This is not how I would have done this.  This is just a matter of style.

This code is in need of a refactor and clean up.  This change would probably be changed when we eventually get to refactoring this code.

PR #2431, opened 2022-07-13
[Scrollabel] revert rename setPos to setPosition

jbleyel:


IanSav (approved):
Good pickup on the conflict with GUIComponent.py.

PR #2433, opened 2022-07-17
[InfoBar] Fix #2414 alphabetic sort of extension menu

### LINKED ISSUE #2414: Sort Extensions not working
https://github.com/openatv/enigma2/blob/7.0/lib/python/Screens/InfoBarGenerics.py#L3161

https://github.com/openatv/enigma2/blob/7.0/lib/python/Screens/InfoBarGenerics.py#L3212

4l3x2k:


4l3x2k:
PR is closed because when too many extensions are installed that do not fit on number keys they are not sorted.

It is even not clear where the sort should happen (see code in #2414).

PR #2434, opened 2022-07-17
[eListbox]

jbleyel:
* add additional moveXX to ListboxActions enum
* reorder ListboxActions
* add goXX functions for movement
* moveToEnd -> Deprecated

### DIFF
--- a/lib/gui/elistbox.cpp
+++ b/lib/gui/elistbox.cpp
@@ -125,8 +125,10 @@ bool eListbox::atEnd()
 	return false;
 }
 
+// Deprecated
 void eListbox::moveToEnd()
 {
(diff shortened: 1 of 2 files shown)


IanSav (approved):
Can you please consider keeping the "goFirst" and "goLast" in the code but commented out to show the reservation of these names for future use when there are official horizontal lists, etc.

The order I have been using is:
goTop()
goPageUp()
goLineUp()
goFirst()
goLeft()
goRight()
goLast()
goLineDown()
goPageDown()
goBottom()

This list order matches how I define all the ActionMaps in the code.  I see this as a logical arrangement when visualising the navigational options.  (Top of the list, page up, line up, first item on the line, previous item on the line, next item on the line, last item on the line, down a line, down a page and finally the bottom of the list.

PR #2442, opened 2022-07-22
Update hints texts from ConfigListScreen

Huevos:
See: https://github.com/OpenViX/enigma2/commit/05b20f3efdcbf371d8dfeaa009ba8acf30f5bf37

This avoids adding pointless code in plugins such as here: https://github.com/oe-alliance/oe-alliance-core/commit/ed1a053bcacb9e17218f3b4cbbed2680ff0d5f5c

Code such as that is not needed in the plugin.

def selectionChanged(self):
    if self["config"]:


Huevos:
Closed and resent here: https://github.com/openatv/enigma2/pull/2443

PR #2443, opened 2022-07-22
ConfigListScreen hints text

Huevos:
This avoids adding pointless code in plugins such as here:
https://github.com/oe-mirrors/e2m3u2bouquet-plugin/commit/6725d52e3a462add46cbe06d58680b55ffff93f0

def selectionChanged(self):
    if self["config"]:

This code should be inherited, not duplicated.

This commit allows the removal of pointless hints text code throughout enigma and in plugins

### DIFF
--- a/lib/python/Screens/Setup.py
+++ b/lib/python/Screens/Setup.py
@@ -213,9 +213,6 @@ class Setup(ConfigListScreen, Screen, HelpableScreen):
 	def selectionChanged(self):
 		if self["config"]:
 			self.setFootnote(None)
-			self["description"].text = self.getCurrentDescription()
-		else:
-			self["description"].text = _("There are no items currently available for this screen.")
 
 	def layoutFinished(self):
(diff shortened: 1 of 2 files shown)


lib/python/Screens/Setup.py
```diff
@@ -213,9 +213,6 @@ def logIncludeElementError(self, element, type, token):
 	def selectionChanged(self):
 		if self["config"]:
 			self.setFootnote(None)
-			self["description"].text = self.getCurrentDescription()
-		else:
```
jbleyel, line 217: Please don't remove the else.

Huevos: The "else" is junk that should not be displayed in the GUI to users. If you don't want the commit please don't merge it.

Huevos:
Ok, it is upgraded as requested.

https://github.com/openatv/enigma2/pull/2444

PR #2444, opened 2022-07-22
ConfigListScreen hints text

Huevos:
This avoids adding pointless code in plugins such as here:
https://github.com/oe-mirrors/e2m3u2bouquet-plugin/commit/6725d52e3a462add46cbe06d58680b55ffff93f0

def selectionChanged(self):
    if self["config"]:

This code should be inherited, not duplicated.

This commit allows the removal of pointless hints text code throughout enigma and in plugins


Huevos:
https://github.com/openatv/enigma2/blob/7.0/lib/python/Screens/Satconfig.py#L793

The commit means stuff such as the above, that is littered around enigma, can now be removed.

PR #2445, opened 2022-07-22
Update it.po

mapi68:


jbleyel:
What should I do with the weblate changes?

PR #2473, opened 2022-08-15
Gif

jbleyel:
[epng] add loadGIF


wedebe:
Almost seems to be a shame to introduce such an old technology at this point in time.

Where will this be used?

IanSav:
Someone has requested access to web page based images.  GIFs are still common on web platforms.  This change will allow any GIFs encountered to also be displayed.

wedebe:
I'm just curious, really. Is it for a plugin (sorry, I can't find anything about the request in the issues page).

I think with `jpg` and `gif` added, `epng.cpp` and `epng.h` should be renamed to `eimg.*`?

jbleyel:
Rename of the file and class can have a lot of side effects.

PR #2517, opened 2022-09-02
[Recoding/Timeshift] setup

jbleyel:
* don't block the setup if the directory not exists


IanSav:
I am awaiting results of discussions amongst the developers.

PR #2535, opened 2022-09-07
[eListboxPythonConfigContent] add support for old skins to calculate …

jbleyel:
…the correct value font based on the entry font if you only have "font" attribute in skin

### DIFF
--- a/lib/gui/elistbox.cpp
+++ b/lib/gui/elistbox.cpp
@@ -506,6 +506,10 @@ void eListbox::entryReset(bool selectionHome)
 void eListbox::setFont(gFont *font)
 {
 	m_style.m_font = font;
+	// support old skins configlist
+	if(!m_style.m_valuefont)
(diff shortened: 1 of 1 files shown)


persianpros:
Our tester confirmed that is ok now, thanks 👍

jbleyel:
I will think about that. Because we don't really need this.
m_valuefont is only needed for configlist and will be set for all list boxes.

persianpros:
We're using new variables everywhere and thanks to https://github.com/OpenVisionE2/enigma2-openvision/blob/develop/CI/skin.sh and https://github.com/OpenVisionE2/enigma2-openvision/blob/develop/CI/pyskin.sh github actions converts any old variable but there are things out there we don't maintain.

For 3rd-party skins I added https://github.com/OpenVisionE2/enigma2-openvision/commit/0d11d99357cd7ec598640003f70413521d54fcba which is ugly now and not yet finished but does the job in less than 3 seconds.

For 3rd-party plugins no solution yet.

PR #2543, opened 2022-09-11
Refactor Timeshift modules and UI 

IanSav:
[Timeshift.py] Update screen code
- Update screen code to latest standards
- No longer block the UI from leaving the "Time shift buffer location" entry or saving the screen if the time shift location is invalid.  Now just warn users they are in for trouble if they don't save reasonable values.
- Use the new DEFAULT_INHIBIT_DEVICES list from LocationBox.py rather than duplicating the effort.
- Simplify the code to reflect the less restrictive UI.
- Ensure the "Time shift buffer location" footnote is correctly displayed when appropriate.

[Timeshift.py] Update the component code
- Merge content of Tools/TimeShift.py into Components/Timeshift.py.
- Clean and sort the imports.
- Remove boxbranding.
- Fix and improve many log comments.
- Don't use int() for integer configuration variables.
- PEP8 clean up the code.

  This code requires further optimization and refactoring.

[setup.xml] Match the "Time Shift" settings to the new code

[UsageConfig.xml] Match the "Time Shift" settings to the new code
- Move all time shift related settings to a single block.  (All such blocks will eventually be sorted into alphabetical order.
- Adjust configuration option names to match the refactored code.

[LocationBox.py] Tidy up code
- Tidy up the code to current standards.
- Add error logs for failed operations.
- Use Booleans rather than 0 and 1 for more clarity.
- Add the more appropriate DEFAULT_INHIBIT_DIRECTORIES name for the old defaultInhibitDirs variable.  This is a global variable and should be shown as such.
- Add the new DEFAULT_INHIBIT_DEVICES list to match an expanded DEFAULT_INHIBIT_DIRECTORIES tuple.

[InfoBarGenerics.py] Use new time shift path

[Migration.py] Tweak the migration process
- Upgrade time shift settings.
- Don't store data in the global data space.
- Only apply each migration step once.
- If the current migration setting can't be determined assume that all migrations are required.
- To remove any item from the settings file set its current value to its default value.
- Small code tidy up.

NOTE: The change in the Time Shift settings UI now allows for invalid time shift buffer locations to be saved.  This change was made to satisfy user requests.  IT IS NOW THE USER'S RESPONSIBILITY TO ENSURE THAT THE TIME SHIFT BUFFER LOCATION IS VALID AND ACCEPTABLE!


IanSav:
SONAR Code smells in Components/Timeshift.py will be addressed when the code is optimised and cleaned up.

PR #2660, opened 2022-11-19
[BoxInfo] add missing grouping

Huevos:


jbleyel:
Are you sure?

```
>>> value
'"TEST"'
>>> value.startswith("\"") or value.startswith("'") and value.endswith(value[0])
True

```

Huevos:
Yes I'm sure. it works different for double quotes than single quotes:

```
>>> value = '"TEST'
>>> value.startswith("\"") or value.startswith("'") and value.endswith(value[0])
True
>>>
>>> value = "'TEST"
>>> value.startswith("\"") or value.startswith("'") and value.endswith(value[0])
False
```
The "and" is to check opening and closing quotes. For double quotes this does not work properly. If the closing quote is missing the current code still returns True.

PR #2662, opened 2022-11-20
[SystemInfo.py] Allow for positive integers to be accepted

IanSav:


IanSav:
This needs to be adjusted and done in two places.

PR #2697, opened 2022-12-04
[FileCommander] improvments 

geeeeeeeeeee:
Fix line selection for move and delete
ReAdd open ViewEdit by startTeletext

### DIFF
--- a/lib/python/Plugins/Extensions/FileCommander/plugin.py
+++ b/lib/python/Plugins/Extensions/FileCommander/plugin.py
@@ -305,7 +305,7 @@ class FileCommander(Screen, HelpableScreen, NumericalTextInput, StatInfo):
 		self["key_blue"] = StaticText()
 		self["key_menu"] = StaticText(_("MENU"))
 		self["key_info"] = StaticText(_("INFO"))
-		self["actions"] = HelpableActionMap(self, ["OkCancelActions", "MenuActions", "InfoActions", "ColorActions", "NavigationActions"], {
+		self["actions"] = HelpableActionMap(self, ["OkCancelActions", "MenuActions", "InfoActions", "ColorActions", "NavigationActions", "InfobarTeletextActions"], {
 			"cancel": (self.keyExit, _("Exit File Commander")),
 			"ok": (self.keyOk, _("Enter a directory or process a file (play/view/edit/install/extract/run)")),
 			"menu": (self.keyMenu, _("Open context menu (contains settings menu)")),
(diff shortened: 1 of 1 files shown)


lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -305,7 +305,7 @@ def __init__(self, session, pathLeft="", pathRight="", leftActive=None):
 		self["key_blue"] = StaticText()
 		self["key_menu"] = StaticText(_("MENU"))
 		self["key_info"] = StaticText(_("INFO"))
-		self["actions"] = HelpableActionMap(self, ["OkCancelActions", "MenuActions", "InfoActions", "ColorActions", "NavigationActions"], {
+		self["actions"] = HelpableActionMap(self, ["OkCancelActions", "MenuActions", "InfoActions", "ColorActions", "NavigationActions", "InfobarTeletextActions"], {
```
jbleyel, line 308: InfobarTeletextActions is not good

lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -325,7 +325,8 @@ def __init__(self, session, pathLeft="", pathRight="", leftActive=None):
 			"right": (self.keyGoRightColumn, _("Switch to the right column")),
 			"down": (self.keyGoLineDown, _("Move down a line")),
 			"pageDown": (self.keyGoPageDown, _("Move down a screen")),
-			"bottom": (self.keyGoBottom, _("Move to last line / screen"))
+			"bottom": (self.keyGoBottom, _("Move to last line / screen")),
+			"startTeletext": (self.keyViewEdit, _("View or edit file (if size < 1MB)"))
```
jbleyel, line 329: There is no protection.
The key is always working and this is not good.

lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -652,6 +653,10 @@ def successCallback(job):
 			if "status" in self:
 				self.displayStatus(_("Delete job completed."))
 				self.sourceColumn.refresh()
+				if startIndex >= self.sourceColumn.count():
+					self.sourceColumn.goBottom()
+				else:
+					self.sourceColumn.setCurrentIndex(startIndex)
```
geeeeeeeeeee, line 659: You can extract this in a method too and rename `startIndex`.

lib/python/Plugins/Extensions/FileCommander/plugin.py
geeeeeeeeeee: I want this key always working but

I'm sorry. Feel free to improve. Maybe i can see and learn.

PR #2711, opened 2022-12-10
[WOL]

jbleyel:
* improve and fix read write WakeOnLAN proc values


IanSav (changes requested):
Please implement the changes we discussed.

IanSav (approved):
Much better.  Easier to read and less to maintain.

PR #2712, opened 2022-12-10
File commander updates

jbleyel:


lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -433,7 +433,7 @@ def updateButtons(self):
 		else:
 			self["key_red"].setText("")
 			self["deleteAction"].setEnabled(False)
-		if currentDirectory and srcPath and self.targetColumn.getPath() and srcName and not srcName.startswith("<") and self.targetColumn.getName():
+		if currentDirectory and srcPath and srcName and not srcName.startswith("<") and self.targetColumn.getCurrentDirectory() and currentDirectory != self.targetColumn.getCurrentDirectory():
```
jbleyel, line 436: Hi @geeeeeeeeeee , you have broken other functions with this change.
Please be careful and make more testing before you make the commit.

Issue #2648, opened 2022-11-07
PluginDescriptor.WHERE_HOTKEY


geeeeeeeeeee:
### German
Ich würde gerne ein (eigenes) Plugin dauerhaft auf einen Hotkey legen.
Dann brauche ich es aber nicht mehr im Erweiterungsmenü etc. aber da beißt sich die Katze...
Nur Plugins die in "PLUGINMENU, EXTENSIONSMENU, EVENTINFO" sind, lassen sich auf einen Hotkey legen.

**Wie wäre es mit einem PluginDescriptor.WHERE_BUTTONSETUP(HOTKEY)?**
Bin da jetzt shon das zweite Mal hängen geblieben, daher die Frage!
ButtonSetupFunctions ist leider auch nicht global/zugänglich.

### English Translator
I would like to put a (own) plugin permanently on a hotkey.
Then I don't need it anymore in the extension menu etc. but that's where the cat bites...
Only plugins that are in "PLUGINMENU, EXTENSIONSMENU, EVENTINFO" can be assigned to a hotkey.

**How about a PluginDescriptor.WHERE_BUTTONSETUP(HOTKEY)?**
I got stuck there for the second time, hence the question!
Unfortunately, ButtonSetupFunctions is also not global/accessible.

geeeeeeeeeee:
Thanks for visit.
But then i don't need it in any menu.
Later I will try to change the code by my self.

jbleyel:
This all sounds like a hack to modify the hotkey functionality and you may break this if you do this in your plugin.
And by the way, what key do you wanna define as hotkey.
Always the same?
How many plugins do you wanna change for this?

geeeeeeeeeee:
Then it's possible to create a second extensionmenu on a hotkey with own plugins that have this descriptor.
And i don't have a plugin in the extensionmenu, that i never call this way, because its on a hotkey.

geeeeeeeeeee:
Then it's possible to create a second "extensionmenu" on a hotkey with own plugins that have this descriptor.
And i don't have a plugin in the extensionmenu, that i never call this way, because its on a hotkey.
I can understand that usually public plugins will never have this desciptor!
It's practically possible to extend this list in a way by a the plugin:
`ButtonSetupFunctions.append((_("Show vertical Program Guide"), "Infobar/openVerticalEPG", "EPG"))
	ButtonSetupFunctions.append((_("Show graphical multi EPG"), "Infobar/openGraphEPG", "EPG"))
[...]

PR #2713, opened 2022-12-10
[ButtonSetup]

jbleyel:
* allow plugin to assign to hotkey only

#2648


geeeeeeeeeee:
I like it.
1st test perfect.

jbleyel:
I have thought about this and the conclusion is: 
This change is very concerning as you are not able to easily remove the plugin as it is not listed.

geeeeeeeeeee:
You mean disable? not remove!
The way to disable is to remove!

jbleyel:
There was a misunderstanding.

PR #2723, opened 2022-12-21
[Input.py] Py3 end() require textU too

geeeeeeeeeee:


IanSav:
What does this change fix?  How do we see the issue?

geeeeeeeeeee:
I have a input and can't use this function to skip to the end, because it don't hit the end, it reaches something before.
This solves the bug. No Issue to show, sorry.
You can close if u like... or merge and solve.

IanSav:
Can you please give me a specific screen where you observe the issue.

IanSav:
Okay, I found the bug you are fixing.

To reproduce the issue edit some text in VirtualKeyBoard and use the button to jump to the end of the text string in the text window.  The cursor will disappear.  The code proposed fixes this issue.

I will include this fix as I perform some other clean ups in this code.

Thank you for your efforts with this bug.  In the future can you please provide more information about bugs and how to reproduce the issue in your contributions.  It took me quite a while to guess and reproduce the issue so that I could understand what you were fixing.

PR #2730, opened 2022-12-26
Update plugin.py

willy014:


jbleyel:
The embedded skin is only for fallback and should be correct.

Issue #2446, opened 2022-07-23
movie/timeshift folders are being created automatically on hdd


Kerberos1864:
Dear Lads,

I noticed something related to openATV 6.4 on Zgemma H9S.
Even with setting Recording folder to /media/mmc/movie and Timeshift folder to /media/mmc/timeshift.
OpenATV is creating these two folder again at each reboot or powercycle in /media/hdd/
How to solve this ?

Best Regards,

wedebe:
Wonder if this is because the mount of `/media/mmc/` hasn't happened before the path check and set to the default of `/media/hdd/`.

I've had empty movie and timeshift folders peppered across my flash and external hdds over time and manually delete them.

ozone1864:
> Wonder if this is because the mount of `/media/mmc/` hasn't happened before the path check and set to the default of `/media/hdd/`.
> 
> I've had empty movie and timeshift folders peppered across my flash and external hdds over time and manually delete them.
> 
> Perhaps (once the mount/create order has been rectified) at the very least a notification to advise that default paths are being used... _ideally_ even with an option to open the `Recording` settings screen.

That's very weird because I already setup manually every single variable related to Recording and Timeshift to /media/mmc.
But still at each reboot two new empty folders are being created in /media/hdd.
Its annoying to delete them every reboot

4l3x2k:
openATV 6.4 won't receive updates anymore.

Please move to openATV 7.2.

I proposed a fix for this issue for openATV 7.2.

AbuBaniaz:
If this issue is fixed on 7.2, can it be closed please?

PR #2747, opened 2023-01-07
[UsageConfig] Enhance creation of movie and time shift directories

4l3x2k:
This fixes Issue #2446.
The default directories for movie and time shift only created if the user specified directories do not exist and fail to be created.


IanSav (commented):
Suggested changes provided to @jbleyel for consideration.

PR #2754, opened 2023-01-16
[FileCommander]

jbleyel:
* add text editor context menu

### DIFF
--- a/lib/python/Plugins/Extensions/FileCommander/plugin.py
+++ b/lib/python/Plugins/Extensions/FileCommander/plugin.py
@@ -2677,7 +2677,7 @@ class FileCommanderTextEditor(Screen, HelpableScreen):
 			self["data"].setList(self.data)
 			self.isChanged = True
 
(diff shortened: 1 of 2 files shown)


lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -2663,6 +2663,29 @@ def keyCancelCallback(self, answer):
...
+				case "SORTA":
+					self.data.sort()
+				case "SORTD":
+					self.data.sort(reverse=True)
+				case "":
+					return
+			self["data"].setList(self.data)
+			self.isChanged = True
+
+		msg = [_("Please select the action for line number %d.") % self["data"].getCurrentIndex() + 1]
```
geeeeeeeeeee, line 2680: msg = [_("Please select the action for line number %d.") % **(**self["data"].getCurrentIndex() + 1**)**]
TypeError: can only concatenate str (not "int") to str

jbleyel:
Hi @geeeeeeeeeee, there is a new FileCommander branch from Ian, please test this new version.

PR #2782, opened 2023-02-12
Unicable- GT Sat

AbuBaniaz:
The "hello" family name is correct and works. Some people like to have the exact model name.


AbuBaniaz:
If accepted, please cherry pick into master/development  branch.

PR #2804, opened 2023-03-25
Cleanup Six

jbleyel:


IanSav (changes requested):
Dictionary items() and values() are already iterable lists so they should need an iter() when used in a loop.

PR #2807, opened 2023-03-29
[AVSwitch]

jbleyel:
* reorg imports
* use "with" for file io


IanSav (changes requested):
I have offered a completely refactored version of this code.  It has been passed to @jbleyel directly and the final version will replace this code when the review is complete.

PR #2839, opened 2023-04-29
[About.py] Remove stray blank line

IanSav:


jbleyel:
This will be done with the next refactor of About.py

PR #2843, opened 2023-04-29
[InfoBarGenerics.py] Optimize the new amlogic code

IanSav:


persianpros:
Nice :+1: 

In OV I use the exact BoxInfo name so instead of "if amlogic" I use "if AmlogicFamily".

BTW ATV image can not be compiled:

image_2023_04_29T08_58_27_846Z-1

PR #2845, opened 2023-04-30
Introduce new eAVControl

jbleyel:
* this new class will be the successor for eAVSwitch
* add several new AV proc functions


persianpros:
Wonderful, more C/C++ code instead of python makes e2 faster and more stable :+1:

jbleyel:
> Wonderful, more C/C++ code instead of python makes e2 faster and more stable 👍
> 
> There are many places in e2 which affect A/V settings and they all seem like duplicates, I hope you eliminate them all and use this new class.

That's the plan. I will move as much as I can to AVControl and remove the duplicate code.

persianpros:
As you're pro with C/C++ can you take a look at https://github.com/openatv/enigma2/blob/master/lib/dvb/tsparser.h#L142 class and see why we're getting

decoder.cpp: error: undefined reference to 'eTsParser::getPTS(long long&)'
decoder.cpp: error: undefined reference to 'eTsParser::flush()'
decoder.cpp: error: undefined reference to 'eTsParser::eTsParser()'
decoder.cpp: error: undefined reference to 'eTsParser::startPid(int)'
decoder.cpp: error: undefined reference to 'eTsParser::stop()'
decoder.cpp: error: undefined reference to 'eTsParser::stop()'
decoder.cpp: error: undefined reference to 'eTsParser::flush()'
decoder.cpp: error: undefined reference to 'eTsParser::freeze()'
decoder.cpp: error: undefined reference to 'eTsParser::unfreeze()'
decoder.cpp: error: undefined reference to 'eTsParser::getPTS(long long&)'
decoder.cpp: error: undefined reference to 'eTsParser::flush()'

errors for https://github.com/openatv/enigma2/blob/master/lib/dvb/decoder.cpp#L67 ?

IanSav:
There was a more significant logic other than performance improvements.  We went for the C++ option as we get to use conditional compilation that is appropriate for *every* box built.  This means that a significant amount of Python code with lots of hardware based "if" logic can be eliminated.  We now get the data needed by Python without having to work out the type of hardware upon which we are running.  The only code being moved to C++ is the code used to directly interface with the hardware.  All the UI and user facing code is still in Python.

wedebe:
Shouldn't such a significant change go into the next version (7.3) instead of master (7.2)?

jbleyel:
> Shouldn't such a significant change go into the next version (7.3) instead of master (7.2)?

The master branch is version 7.3.
This means all changes in master will NOT come in 7.2 except small changes or bugfixes.

persianpros:
I think https://github.com/openatv/enigma2/blob/eAVControl/lib/driver/avcontrol.cpp#L121 is wrong and should be "framerate"

jbleyel:
> I think https://github.com/openatv/enigma2/blob/eAVControl/lib/driver/avcontrol.cpp#L121 is wrong and should be "framerate"

Fixed .. thanks

persianpros:
Seems parts of the code is not compatible with newer C++ which we're using:

`error: no match for 'operator=' (operand types are 'const std::string' {aka 'const std::__cxx11::basic_string<char>'} and 'std::string' {aka 'std::__cxx11::basic_string<char>'})
  203 |                 m_video_mode_60 = CFile::read(proc_videomode_60);
      |                                                                ^`

jbleyel:
Hi @persianpros , please feel free and tell me what function is missing in avcontrol.

persianpros:
Sure, I have to check your file again (with latest updates) first. Hope we could compile it with C++ 17 also otherwise we can't use it in OV.

Don't we need /proc/stb/video/videomode_24hz? We're checking it in Wizard and AV settings.

I think the naming of this file is not correct as it's about Video mostly, better to use VideoControl and have another one for Audio like AudioControl and manage AudioSelection and AVSwitch related things.

Even for debugging and maintenance it would be easier to have two parts instead of one with lots of checks, as you're a pro with C you could also add another one for LCD/LED/VFD related things.

A new folder for this matter would be even more nice, I mean instead of /lib/driver use /lib/helper

Great job like always.

jbleyel:
> Sure, I have to check your file again (with latest updates) first. Hope we could compile it with C++ 17 also otherwise we can't use it in OV.
> 
> Don't we need /proc/stb/video/videomode_24hz? We're checking it in Wizard and AV settings.
> 
> I think the naming of this file is not correct as it's about Video mostly, better to use VideoControl and have another one for Audio like AudioControl and manage AudioSelection and AVSwitch related things.
> 
> Even for debugging and maintenance it would be easier to have two parts instead of one with lots of checks, as you're a pro with C you could also add another one for LCD/LED/VFD related things.
> 
> A new folder for this matter would be even more nice, I mean instead of /lib/driver use /lib/helper
> 
> Great job like always.

AV means Audio/Video .. I don't wanna make 2 files.
LCD is separated already.

What exactly is the problem of C++ 17?

PR #2863, opened 2023-05-04
LCD Improvements

jbleyel:


persianpros:
I think you need

`#define LCD_DM9X0_Y_OFFSET 4`

in lcd.cpp

jbleyel:
> I think you need
> 
> `#define LCD_DM900_Y_OFFSET 4`
> 
> in lcd.cpp

That’s defined in configure.ac

persianpros:
Ah ok not a bad idea to move it there, I think we can clean more like this way as there are more things like that.

PR #2874, opened 2023-05-09
[Components/Console] Fix 'Console' object has no attribute 'appResults'

m4dhouse:


jbleyel:
Where is this "self.appResults" used?
It's not used inside of Console.py

PR #2899, opened 2023-05-20
7.2 initialize format

AbuBaniaz:


jbleyel:
We will do this within the refactor of these files.

PR #2919, opened 2023-06-01
[Screen.py] fix incorrect selection of summary screen

Huevos:


IanSav:
Where have you seen an issue?  What happened in that case?

Huevos:
So Ian, the current code could work unexpectedly in lots of places, I've seen it in several screens, but to give you a concrete example...

Without the patch:
lcdshot_20230601174246

With the patch:
lcdshot_20230602002610

Huevos:
So typically the summary screen is named after the parent class, e.g. Menu -> MenuSummary, but that is not a rule. Sometimes the summary class being used has a different name. Maybe it is even inherited. So it makes sense to check the the name of the summary class in use before falling back to ScreenSummary/SimpleSummary.

AbuBaniaz:
Another example is Cablescan

PR #2923, opened 2023-06-04
PEP8

persianpros:


jbleyel:
I don’t like this.
We will refactor each file step by step and then we also make PEP8.

PR #2928, opened 2023-06-06
Added a cache for skin filenames to improve EPG performance.

dachaplin:
I've been profiling opening the graphical EPG in MetrixHD (on a Zgemma H.2S) and noticed that a lot of time is spent in resolveFilename():

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    1.776    1.776 /usr/lib/enigma2/python/Screens/InfoBarGenerics.py:2021(openGraphEPG)
        ...
       75    0.080    0.001    0.432    0.006 /usr/lib/enigma2/python/Tools/Directories.py:90(resolveFilename)

```
It is taking 0.432 seconds just to resolve filenames. With more rows in the EPG and using other skins this can go even higher (it was over 0.5 seconds for me, using Slyk Onyx with 16 rows).

I tried adding a cache of resolved filenames and this reduced the time considerably, especially once the cache is created after the first call to openGraphEPG():

```
       75    0.049    0.001    0.214    0.003 /usr/lib/enigma2/python/Tools/Directories.py:91(resolveFilename)

       75    0.007    0.000    0.008    0.000 /usr/lib/enigma2/python/Tools/Directories.py:91(resolveFilename)

       75    0.007    0.000    0.008    0.000 /usr/lib/enigma2/python/Tools/Directories.py:91(resolveFilename)

       75    0.006    0.000    0.007    0.000 /usr/lib/enigma2/python/Tools/Directories.py:91(resolveFilename)

```
I'm new to the Enigma code so I'm not sure if this will cause problems elsewhere, but hopefully something like this can be done.


jbleyel:
resolveFilename is used all over the place and not only in graphical EPG.
This means you will cache all images and that's not acceptable.

If you wanna have cache the you should cache the image itself in graphical EPG and not the filename.

dachaplin:
I restricted it to SCOPE_GUISKIN so it only caches the filenames of skin files. Skin files don't move around, do they?

Taking 0.4-0.5 seconds just to find these files seems ridiculous.

jbleyel:
You probably miss one thing.
We also support other skins like Metrix and resolveFilename is also for fallback if the image is not in the main skin.
As I said .. You really wanna cache ALL image file names in all screens?
For Metrix : we talk about round about 1560 files + all files from default skin which are not in Metrix

dachaplin:
Maybe just turn caching on in places like opening the EPG, e.g. setResolveFilenameCaching(true)

jbleyel:
Can you please run your box for a while and open a lot of screens and plugins and then dump the content of your cache in the log or extra file.
You should call the dump in StartEnigma after that line 

https://github.com/openatv/enigma2/blob/master/lib/python/StartEnigma.py#L938

Restart GUI will then call the dump

jbleyel:
Hi @dachaplin, we will do that differently by adding the cache into EPG.
Your finding is correct but your suggestion can be more optimized by caching the images itself and not only the names.
This improvement is already in our pipeline.
So please be patient.

dachaplin:
OK. Let me know if you still want the cache dump.

jbleyel:
You can post the dump. This will help us to find the files.

kiddac:
There is probably no way of knowing if a skin or plugin is programmatically changing the same image. 
Caching images will cause problems for any skin or plugin that use the same named graphic file for picons, posters, backdrops or any other changed graphic. 
Don't do this.

jbleyel:
Can you please make a new measurement using the latest 7.6 code because we have changed resolveFilename.

dachaplin:
I'm afraid I'm not using OpenATV any more - my satellite dish is currently broken!

jbleyel:
> I'm afraid I'm not using OpenATV any more - my satellite dish is currently broken!

But you may able to provide the info how do you have done the messurement.

dachaplin:
I think I used cProfile, like the commented-out code below. I might have wrapped openGraphEPG() instead though.


import cProfile, pstats, io
from pstats import SortKey

			#pr = cProfile.Profile()
			#pr.enable()

			self.showBouquetNumber(number)

			#pr.disable()
			#s = io.StringIO()
			#sortby = SortKey.CUMULATIVE
			#ps = pstats.Stats(pr, stream=s).sort_stats(sortby)
			#ps.print_stats()
			#print(s.getvalue())

PR #2948, opened 2023-06-17
Changed graphical EPG to only load event data for visible channels.

dachaplin:
Currently the Graphical EPG is very slow for bouquets with lots of channels (e.g. the "Last Scanned" bouquet).
Moving to the bouquet, and stepping forward or backward in time takes several seconds. (I'm using a Zgemma H.2S.)

The problem seems to be that the Graphical EPG code is loading EPG data for all the channels in the bouquet at once.
This patch changes it so it only loads data for the channels visible in the EPG. (It only keeps visible channels in the list.)
It is much faster for me now.

I'm still new to the OpenATV code, so there are probably things I've missed (e.g. certain config options may be broken). But I think something like this needs to be done to speed up the EPG.

### DIFF
--- a/lib/python/Components/EpgList.py
+++ b/lib/python/Components/EpgList.py
@@ -81,6 +81,10 @@ class EPGList(GUIComponent):
 		self.days = (_("Mon"), _("Tue"), _("Wed"), _("Thu"), _("Fri"), _("Sat"), _("Sun"))
 		#//
 
+		self.listRows = 8
+		self.listFirstServiceIndex = 0
+		self.serviceList = ()
+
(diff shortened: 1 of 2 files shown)


dachaplin:
It calls queryEPG() each time though. It shouldn't really need to do that twice - it seems a waste of time.

jbleyel:
Do you have also tested the other EPG modes with your changes?

dachaplin:
Yes, I've had a look at them and didn't notice any problems.

GhostofGeeeee:
@dachaplin
We detect some side effects.

PR #2949, opened 2023-06-17
Unicable connected

AbuBaniaz:
Include wording that is in hint text. It is not asking if you have a signal wire `connected` to the tuner, it is as asking if you are `connected through another tuner`.  Make text reflect the question better.

Before:
ATV- Connnected-orig

After:
ATV- Connnected-modded


jbleyel:
We can do that, but the whole file needs a refactor.

PR #2960, opened 2023-06-28
[eRectangle]

jbleyel:
* remove setBackgroundGradient

[eWidget]
* add setBackgroundGradient to eWidget

[Skin]
* fix parseGradient
* fix backgroundGradient attribute names


persianpros:
Not sure what's wrong with our code but https://github.com/OpenVisionE2/enigma2-openvision/commit/fb0a69418e43245f5da71b86be33b72d74d10cf2 is not compatible with OV and I had to revert it sadly.

We only see the first line in every menu and the rest are hidden.

PR #2973, opened 2023-07-11
[LocaleSelection]

jbleyel:
* add restart question after language change.

### DIFF
--- a/lib/python/Screens/LocaleSelection.py
+++ b/lib/python/Screens/LocaleSelection.py
@@ -11,6 +11,7 @@ from Screens.HelpMenu import HelpableScreen, ShowRemoteControl
 from Screens.MessageBox import MessageBox
 from Screens.Screen import Screen, ScreenSummary
 from Screens.Setup import Setup
+from Screens.Standby import QUIT_RESTART, TryQuitMainloop
 from Tools.Directories import SCOPE_GUISKIN, resolveFilename
 from Tools.LoadPixmap import LoadPixmap
 
@@ -27,8 +28,6 @@ config.locales.localesSortBy = ConfigSelection(default="2", choices=[
 	("3", _("Locale (Ascending)")),
 	("30", _("Locale (Descending)"))
 ])
-
-inWizard = False
 
 
 class LocaleSelection(Screen, HelpableScreen):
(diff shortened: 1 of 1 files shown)


atvcaptain:
fix the problem with mixing languages after the switch, it makes sense to add this

IanSav:
How often do users change the language of a running openATV installation?  Is this often enough to warrant this extra code and pop up box?  I suspect that the answer will be that this is not important to be done here.

I also suspect that any user's who loose a recording due to this reboot will be VERY unhappy.  If they don't allow the reboot then the problem remains.  This solution is not very comprehensive and will still allow issue.  Is this worth it?

This almost falls into the category of unneeded bloat.  I suspect that @jbleyel's fixes to the Wizard will be more than enough to solve the language issue in its entirety.

atvcaptain (commented):
Users report that because they then say the language does not fit, then they rather take openvix or openpli, because there everything fits, have the info also only from other forums

IanSav:
What do you mean by "fits"?

atvcaptain:
users report that they have a language mix and do not understand that they always have to do a gui restart after changing the language

IanSav:
Is this still an issue with the new Wizard?

Are the users really changing languages all the time?  We should be asking users *WHY* do they need to keep changing languages?  Perhaps they have another issue and they are trying to fix it by changing languages.  We should be focussing on fixing the *REAL* problem.

atvcaptain:
with the new wizard everything is fine, just change the language in the running image, a simple restart of the user interface solves the problem yet

IanSav:
I still don't understand why users are constantly changing languages?  As a developer I change between English and German to work with @jbleyel.  As a user I selected English Australian on the first day and NEVER changed it since.

It seems to me that a developer might be complaining but I doubt that any users would have any issues.  @jbleyel is going to talk to you to find out more about the reasoning and logic behind the request.

I am just worried that this code change is a patch that is hiding the real issue.  It is that real issue I want to understand.  When I wrote to International module I created a callback so that any code can be notified when a locale change is requested.  I would like to know if the issue you are trying to fix would be better suited to getting a direct callback when the locale gets changed rather than forcing a restart that may interrupt active timers.

PR #2979, opened 2023-07-23
Small tweaks to BackupRestore.py

neoatomic:
Small tweaks to BackupRestore.py

Included for backup :

/etc/sysctl.conf
(for users who disable ipv6 its needed to backup.)

Removed for backup :

"/home/"
(Don't see a reason to include the main folder instead of individual files?)

Also settings-backup is getting pretty large over time. Main reason are the EPG cache files.

/etc/enigma2/epg.dat
/etc/enigma2/xstreamity/epg/

Do they really need to be backup'd by default for all users or better to exclude epg? Backup seize are then reduced from ~50mb to ~1mb and much quicker to make and restore backups. Big improvement.


Papi2000-new:
For me, it's OK to add

/etc/sysctl.conf

but the rest have to stay in the defaults!

jbleyel:
Hi @neoatomic , "/etc/sysctl.conf" is OK .. "/home/" is not OK.
Please make a new pull and please pull to master and not to 7.3 branch.

jbleyel:
I'm not sure about 

```
/etc/enigma2/epg.dat
/etc/enigma2/xstreamity/epg/
```

This should be discussed first.

PR #3024, opened 2023-08-23
Cleanup

jbleyel:


IanSav (approved):
Looks good and should speed up the code.

Note to others, only do this for Python 3.

IanSav (commented):
Does this offer a performance improvement?  I am not sure it is as readable.

IanSav (commented):
Does this offer a performance improvement?  I am not sure it is as readable.

IanSav (changes requested):
Is this really equivalent code syntax?  I am thinking that set() is a unique list of items while {} is a dictionary and not a list!

PR #3042, opened 2023-09-06
fix GS in Timers.py

j00zek:
This change fixes GS when getLocation() invoked from line 1524 "self.session.openWithCallback(self.getLocation, MessageBox, _("Error: A location to save the recording has not been selected!"), MessageBox.TYPE_ERROR)"


j00zek:
fix for
10:26:08.0936 TypeError: RecordTimerEdit.getLocation() takes 1 positional argument but 2 were given

PR #3055, opened 2023-09-23
m4/ax_python_devel.m4 - update to #serial 36 providing sysutils suppo…

TwolDE2:
…rt for OE-A 5.3 cross compile libraries, python 3.11 and hopefully python 3.12

### DIFF
--- a/m4/ax_python_devel.m4
+++ b/m4/ax_python_devel.m4
@@ -4,7 +4,7 @@
 #
 # SYNOPSIS
 #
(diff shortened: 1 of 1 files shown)


atvcaptain:
we have merge to master branch if all ok we merge to 7.3 too

TwolDE2:
I built openatv  on master branch so should not be an issue

atvcaptain:
yes all fine np

PR #3067, opened 2023-10-01
[Skin.py] Fix name 'n' is not defined

m4dhouse:


IanSav:
Thanks for spotting and fixing this error.

PR #3103, opened 2023-10-31
Refactor International and LocaleSelection

IanSav:
[International.py]
- Remove large dictionaries from the global variable space and move them to the International class.
- Remove the repeated "UTF-8" text and use "" to represent UTF-8.  Only specify encodings other than UTF-8.
- Add a new field to specify the default keyboard mappings for each language.  Only specify keyboards other than "qwerty".
- Remove the buildISO3166() method as it and the associated module CountryCodes.py file are no longer required.
- Use f-strings.
- Re-sequence some of the methods and make some small optimizations of the code.
- Remove the deleteLanguagePackages(), installLanguagePackages() and runPackageManager() methods and shift the functionality to LocaleSelection.py where a better interaction with the UI can be created.

[LocaleSelection.py]
- Use the streamlined International module import.
- Remove unused imports.
- Implement methods to add and remove locales / languages.  These replace the methods previously in International.py.
- Use Opkg.py rather than directly accessing the opkg command, this no longer locks the UI while processing.
- Use the Processing screen to show users when there is background processing to add or remove locales / languages.
- Re-sequence some of the methods and make some small optimizations of the code.
- Use "country" and "language" variable names to make the code easier to read.
- Restore the Summary screen that got lost along the way.
- Remove the unused selectKeys() method.

[CountryCodes.py]
- Module removed as it is no longer needed.


IanSav:
This pull request has been superseded.

PR #3118, opened 2023-11-07
GUIAddon

jbleyel:
* implementation of Pager, ButtonSequence and ColorButtonsSequence

This is taken from openVix and needed for skin compatibility reasons.

Co-authored-by: @DimitarCC
Co-authored-by: @Huevos

### DIFF
--- a/lib/python/Components/Converter/Converter.py
+++ b/lib/python/Components/Converter/Converter.py
@@ -5,9 +5,13 @@ class Converter(Element):
 	def __init__(self, arguments):
 		Element.__init__(self)
 		self.converter_arguments = arguments
+		self.separatorChar = ""
 
 	def __repr__(self):
 		return "%s(%s)" % (str(type(self)), self.converter_arguments)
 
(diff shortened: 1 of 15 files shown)


lib/python/Components/Converter/Converter.py
```diff
@@ -5,9 +5,13 @@ class Converter(Element):
...
+		self.separatorChar = ""
 
 	def __repr__(self):
 		return "%s(%s)" % (str(type(self)), self.converter_arguments)
 
 	def handleCommand(self, cmd):
 		self.source.handleCommand(cmd)
+
+	def appendToStringWithSeparator(self, str, part):
+		return part if str == "" else f"{str}{self.separatorChar}{part}"
```
DimitarCC, line 17: Here you have to add spaces around separator since skin engine strips them if you define them in convertor parameter

DimitarCC: There is new implementation of that. You can check it out.

PR #3146, opened 2023-11-22
[ChoiceBox] Make config.misc.pluginlist.fc_bookmarks_order proof

GhostofGeeeee:
config.misc.pluginlist.fc_bookmarks_order couldn't handle bookmarks with "," in path.

- remove config.misc.pluginlist = ConfigSubsection()
- remove config.misc.pluginlist.eventinfo_order
- remove config.misc.pluginlist.extension_order
- remove config.misc.pluginlist.fc_bookmarks_order
- create ConfigArray
- replace reorderConfig
- add new config.plugins.FileCommander.bookmarks_order


jbleyel:
This is not working because ConfigSet has sorting and this will break the whole feature.

GhostofGeeeee:
LOL I see that now too. First test works... I like that.
ConfigLocations is wrong too.
Then I see only a different seperator?

jbleyel:
There is actually no fitting ConfigXX class to support this.
I already told you.. NOT that easy.

jbleyel:
ConfigList class already exists and that's why we should not use this name.
What about ConfigArray?

jbleyel:
AND .. you cannot use ConfigSet as base class for this.

jbleyel:
You can try this.

```
# This is the control, and base class, for a list of items.
#
class ConfigArray(ConfigElement):
	def __init__(self, choices, default=None, visible_width=False):
		ConfigElement.__init__(self)
		if isinstance(choices, list):
			self.choices = choicesList(choices, choicesList.TYPE_LIST)
		else:
			raise TypeError("[Config] Error: 'ConfigArray' choices must be a list!")
		if default is None:
			default = []
		self.default = default
		self.lastValue = default
		self.visibleWidth = visible_width
		self.value = shallowcopy(default)
		self.item = 0

	def load(self):
		ConfigElement.load(self)
		if not isinstance(self.value, list):
			self.value = [] if self.value is None else list(self.value)

	def handleKey(self, key, callback=None):
		count = len(self.value) - 1
		if key == ACTIONKEY_FIRST:
			self.item = 0
		elif key == ACTIONKEY_LEFT:
			self.item = self.item - 1 if self.item > 0 else count
		elif key == ACTIONKEY_RIGHT:
			self.item = self.item + 1 if self.item < count else 0
		elif key == ACTIONKEY_LAST:
			self.item = count

	def getText(self):
		return " ".join(self.value)

	def getMulti(self, selected):
		if selected:
			start = 0
			for index, value in enumerate(self.value):
				if index == self.item:
					break
				start += len(value) + 1
			end = start + len(value)
			value = " ".join(self.value)
			if self.visibleWidth and len(value) > self.visibleWidth:
				if start + 1 < self.visibleWidth / 2:
					offset = 0
				else:
					offset = min(start + 1 - self.visibleWidth / 2, len(value) - self.visibleWidth)
				return ("mtext", value[offset:offset + self.visibleWidth], list(range(start - offset, end - offset)))
			else:
				return ("mtext", value, list(range(start, end)))
		else:
			value = " ".join(self.value)
			return ("text", value[0:self.visibleWidth] if self.visibleWidth and len(value) > self.visibleWidth else value)

	def fromString(self, value):
		return eval(value)

	def toString(self, value):
		return str([x[0] if isinstance(x, list) else x for x in value])

```

jbleyel:
We will not merge this PR.
The new ConfigArray class is not necessary if you format the string correctly.
In addition, a comma in a directory name is not a good idea and therefore it will not be implemented.

jbleyel:
You should also use this new version as starting point.

https://github.com/openatv/enigma2/commit/349e5580e7da3c635b84837649c8c03a8bb1ec8d

GhostofGeeeee:
I think ConfigText isn't good for a list.
The [ChoiceBox]-Branch isn't uptodate with FileCommander...

jbleyel:
No .. The biggest question is ..
Why do you need , in the directory name.
What's the exact purpose?

GhostofGeeeee:
I have a dir "1,2TB" and want this as a bookmark.
A workaround is no problem but why?
All my OS support "," in a path.

jbleyel:
> I have a dir "1,2TB" and want this as a bookmark. 

Why? I don't really understand the reason.

> A workaround is no problem but why? 

An extra class and su much extra effort for a single user problem is priority Z.

> All my OS support "," in a path.

Yes , but this means not automatically that you should use this 

**Is this really needed???**

GhostofGeeeee:
I think this is better.
I don't like to say please but else I need to think about my dir "1,2TB" ;(
And i'm think about searchepg.history unsorted.
I would create ConfigSet as a subclass of ConfigList/Array...

jbleyel:
> I think this is better. I don't like to say please but else I need to think about rename my dir "1,2TB" ;( And i'm think about searchepg.history unsorted. I would create ConfigSet as a subclass of ConfigList/Array...

searchepg.history ? What's this?

GhostofGeeeee:
config.plugins.epgsearch.history = ConfigSet(choices=[])
A sorted history... LOL

jbleyel:
> config.plugins.epgsearch.history = ConfigSet(choices=[]) A sorted history... LOL

This is not part of enigma core.

GhostofGeeeee:
FileCommander is a import too.
Maybe with old dead code
```

config.misc.pluginlist.eventinfo_order = ConfigText(default="")
config.misc.pluginlist.extension_order = ConfigText(default="")
```

jbleyel:
> FileCommander is a import too. Maybe with old dead code
> 
> ```
> 
> config.misc.pluginlist.eventinfo_order = ConfigText(default="")
> config.misc.pluginlist.extension_order = ConfigText(default="")
> ```

No .. this is not fully implemented.
Please don't touch.

GhostofGeeeee:
OK

GhostofGeeeee:
And on 6.4 the sort doesnt work thats why nobody had a problem on it!

jbleyel:
We will refactor ChoiceBox and we may come back to this issue.
But this is not prio because there are a lot of real issues and many other things on our todo list.

GhostofGeeeee:
OK

PR #3179, opened 2023-12-06
text changes

GhostofGeeeee:


lib/python/Screens/InfoBarGenerics.py
```diff
@@ -3913,7 +3913,7 @@ def startInstantRecording(self, limitEvent=False):
 				end = info["end"]
 		else:
 			if limitEvent:
-				self.session.open(MessageBox, _("No event info found, recording indefinitely."), MessageBox.TYPE_INFO)
+				self.session.open(MessageBox, _("No event info found, recording default is indefinitely."), MessageBox.TYPE_INFO)
```
wedebe, line 3916: "Recording indefinitely" is grammatically correct here, albeit functionally inaccurate, as "indefinite" in this case actually meeans (from what I can recall) 24 hours.

PR #3189, opened 2023-12-11
Tweak bsod message text

wedebe:
This change tweaks the crash message text (specifically to update file name format).


jbleyel:
This is only one half.

https://github.com/openatv/enigma2/blob/master/main/bsod.cpp#L278-L303

wedebe:
> This is only one half.
> 
> https://github.com/openatv/enigma2/blob/master/main/bsod.cpp#L278-L303

I'm usually reluctant to change C++ code 'cos I've no way to test my changes. Is it ok with you for me to submit ui-text-only C++ changes?

jbleyel:
I think we should remove the python part for this and do this only in c.

PR #3195, opened 2023-12-15
[Wizard] Change default option for channel list installation to "none"

AbuBaniaz:
Do not install defaultsat (13e/19e) channel lists unless user asks for it.


atvcaptain:
the default setting is not changed because it is an important function, the list is created exclusively for us, that is a lot of work.

PR #3152, opened 2023-11-25
[Instant Record] Improvements + fix setEndime+changeDuration

GhostofGeeeee:
- `setEndime` date was always today
too, on a event which ends tomorrow. --> TimeDateInput refactory

- Prepare for https://github.com/GhostofGeeeee/enigma2/tree/TimeDateInput refactory

- `changeDuration` needs `marginAfter = 0` too.

- Abort `setEndtime` on creation should be indefinitely

- Focus on Clock `dlg.keyDown()`

Please feel free to help me proper the commit documetaion

PR #3198, opened 2023-12-19
[DateTimeInput.py] Replace TimeDateInput with DateTimeInput

IanSav:
[DateTimeInput.py]
- This is a new module that offers a more concise user interface for setting times for instant recordings and EPG navigation.
- This new code is a subclass of Setup to reduce the amount of code copied from Setup.py and ConfigList.py.
- This code uses consistent navigation relative to other Setup based screens.
- The times entered now directly and accurately update the date *without* requiring an extra control with which users must interact.
- The range of times and dates the user may select is now limited to only valid values.  For example, instant recordings are now correctly limited to 24 hours and EPG jumps can be made forward in time as well as backwards in time to the extent of the EPG history that is enabled.
- HELP is now available in this screen.
- Button labels better reflect the actions to be performed when pressed.

[InfoBarGenerics.py]
- Switch from using TimeDateInput.py to DateTimeInput.py.
- Optimize some code.
- Improve some translation strings.
- Use f-strings to improve performance of the button press logging.

[EpgSelection.py]
- Switch from using TimeDateInput.py to DateTimeInput.py.
- Remove use of getSkinFactor as it is not used in openATV (openATV has fully integrated and automatic skin scaling)  and can cause skin issues.
- PEP8 correct some code indentation.
- Improve some translation strings.


GhostofGeeeee:
[InfoBarGenerics.py]

- Discret refere to my [fixes](https://github.com/openatv/enigma2/pull/3152).

jbleyel:
> [InfoBarGenerics.py]
> 
> * Discreet refer to my [fixes](https://github.com/openatv/enigma2/pull/3152).

The new code is different.
Please test.

PR #3236, opened 2024-01-16
[L10n] Update International.py: "ราชอาณาจักรไทย" -> "ประเทศไทย"

bact:
The translation of "Thailand" should be "ประเทศไทย". This is a common form of the country name.

The current translation "ราชอาณาจักรไทย" is only use for the official full form "Kingdom of Thailand".


IanSav:
I am okay with this change.  I was limited by what I could find on Google when I created the original data.

PR #3272, opened 2024-03-11
[skin.py] Correct method spacing

IanSav:


jbleyel:
This will be done in the next change of the file.

PR #3283, opened 2024-03-18
Update Televes and added GT-dLNB1-T - Poland

zukon:


jbleyel:
Please remove the extra spaces of the xml attributes what you have added.

jbleyel:
The format attribute should be always "JESS" not "jess"

PR #3287, opened 2024-04-09
[VirtualKeyBoard.py] Add TAB editing and update keyboard

IanSav:
- Add the ability to edit lines containing a TAB character.
- Start transformation of VirtualKeyBoard to VirtualKeyboard.  (The "B" is now lower case.  This better matches the use of Keyboard elsewhere in Enigma2.)
- Remap some onscreen buttons to better match real keyboards.
  - The FIRST icon location is now the TAB icon.
  - The LAST icon location is now the CAPSLOCK icon.
  - The CAPSLOCK icon locations are now SHIFT icons.
  - The EXIT icon has moved to the right of the SPACE bar where the SHIFT icon used to be located.
  - FIRST icon and LAST icon are now before and after the LEFT and RIGHT icons respectively.
- The color button functions are unchanged.

Coders are asked to please update their code to call the new VirtualKeyboard class and note the variable name change.

Skinners are asked to please update their skins to rename the "VirtualKeyBoard" screen to "VirtualKeyboard".  Also the "VirtualKeyBoard" based font and parameters should be updated to the new "VirtualKeyboard" names.  The current code supports both version of the names but to improve performance we would like to remove the dual name support ASAP.


Ev0-BH:
What is the point of this ?

Just loads of work for plugin developers and skinners for ZERO gain

IanSav:
That is your opinion and you are welcome to have it.

The change was triggered by bug reports where users could not edit files containing TAB characters.  In order to add a TAB button a button space needed to be created.  This necessitated some button rearrangements to preserve the usability of the layout.

Following openATV coding preferences, when old code is revisited it is brought up to the coding standards of the day.  This is an opportunity to tweak and improve the code.  In this case, the keyboard layout has been adjusted to better match the physical keyboards against which this screen was designed.

What effort is required by plugin developers and/or skinners?  The code should be drop in compatible with *all* existing code and skins.  A request has been made for openATV developers and skinners to adjust their code so that the compatibility code can be removed in due course.  (Internal openATV coding changes will be made to use the updated code.  This will have no impact on plugin developers and/or skinners.)

By the way, the "loads of work" took me less than 30 minutes to design and code the changes.  This was one of the fastest and easiest changes I have made.

ian10951:
This is yet another totally pointless commit causing work and grief fo others and for what ?

Its almost like some people on here wake up in a morning and think to themselves what can I do today thats very 
least effort for me but will cause the most disruption possible to others.

I think there will be many complaints about this once it comes into play.

IanSav:
What exactly has changed for you that is causing such feedback?  What is the grief of which I am being accused of creating?  The code was tested to ensure that all existing code and skins would operate without any need for change!  If there is an issue please point it out.  Constructive criticism is welcome, pointless whinging is not.  If there is a bug or problem then please point it out.  Where have I made the "B' to "b" change mandatory?  It is a suggested change but has NOT been enforced anywhere for which I am aware?

I am gratified that you don't want to edit files with TAB characters but your opinion is not shared by others who requested the change.

ian10951:
Quote

 The current code supports both version of the names but to improve performance we would like to remove the dual name support ASAP. 

So soon you suggest it will be manditory.

jbleyel:
@ian10951 , @Ev0-BH : 

Do you mean only the last sentence of the commit note or what else is a problem?

ian10951:
This

Start transformation of VirtualKeyBoard to VirtualKeyboard. (The "B" is now lower case. This better matches the use of Keyboard elsewhere in Enigma2.) 

Its this silly little change thats going to cause a headache for many users. Its this I'm asking if you dont mind reverting ?

Ev0-BH:
> @ian10951 , @Ev0-BH :
> 
> Do you mean only the last sentence of the commit note or what else is a problem?

It is the renaming of VirtualKeyBoard to VirtualKeyboard that is (going to be) the issue.

@IanSav as above I am not complaining about the TAB or the layout changes, it's the planned force change of naming the screen that will cause others (not yourself) to waste lots of time, updating plugins and skins for zero gain, and if / when they do, it will mean all other images will have to also change this, otherwise the plugins (and or skins) won't be compatible with other images again.

You don't seem to realise (or care) that many users use plugins, not all of which are public, not all of which the authors are still active (or even alive), are used to using these things, they have used them for many years and your pointless renaming of things break this. It might be little work for you and take you just a few seconds or minutes to do, it's the stuff that gets broken because of these little changes you make that is the problem and you seem oblivious to it and your attitude is well they should just update it, but for the reasons I pointed out already that isn't always possible.

IanSav:
> It is the renaming of VirtualKeyBoard to VirtualKeyboard that is (going to be) the issue.

How is this change an issue?  What has been broken?  I am yet to hear about any real issues created by these changes.  If there was any credible evidence of an issue I would react very differently to the comments.

As for the removal of the compatibility code, that will not be happening any time soon.

Ev0-BH:
> > It is the renaming of VirtualKeyBoard to VirtualKeyboard that is (going to be) the issue.
> 
> How is this change an issue? What has been broken? I am yet to hear about any real issues created by these changes. If there was any credible evidence of an issue I would react very differently to the comments.
> 
> As for the removal of the compatibility code, that will not be happening any time soon.

Do you not read ?

It is the fact that you plan to remove the compatibility code that is THE ISSUE

ian10951:
> As for the removal of the compatibility code, that will not be happening any time soon.

That contradicts what you said earlier, whereby you said you wanted the old code removed ASAP.

But if this is now the case and the compatibility code is to remain, then there are no issues now at all. 

Thank You for keeping it in place.

IanSav:
When the changes were conceived that was the plan.  The plan has changed based on the feedback.

I should also note that there is a lot of development happening in many areas of the code.  Sometimes the commit notes are rushed and do not come out as intended.  This was one of those cases.

PR #3309, opened 2024-05-04
[ChannelSelection] correct bouquet name for EPG

### LINKED ISSUE #2717: [BUG] Full EPG - Wrong bouquet name when user zaps to a non-bouquet channel
Recreate:
- zap to a channel in (eg.) `Favourites` bouquet
- open Full EPG

Observe:
- title says `<    Favourites    >`
- channel list is correct for contents and order

Now:
- close Full EPG
- open `Channel List` (using `UP` or `DOWN` button
- press `RED` ('All')
- zap to a channel that's not in a bouquet
- open Full EPG

Issue:
- EPG title has kept name of bouquet that was previously selected (eg. in this case: `<    Favourites    >`), but channel list is seemingly random.
- Pressing `Previous` button then goes to relevant bouquet with correct title and correct list of channels

Resolution:
- an appropriate `<    Title    >`

GhostofGeeeee:
for non-bouquet channel
- fix #2717

### DIFF
--- a/lib/python/Screens/ChannelSelection.py
+++ b/lib/python/Screens/ChannelSelection.py
@@ -634,8 +634,6 @@ class ChannelSelectionBase(Screen, HelpableScreen):
 
 	def getBouquetList(self):
 		bouquets = []
-		if self.isSubservices():
- 			bouquets.append((self.getServiceName(self.subservicesBouquet), self.subservicesBouquet))
 		serviceHandler = eServiceCenter.getInstance()
 		if config.usage.multibouquet.value:
 			list = serviceHandler.list(self.bouquet_root)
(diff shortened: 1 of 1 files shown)


lib/python/Screens/ChannelSelection.py
```diff
@@ -634,8 +634,6 @@ def setCurrentSelectionAlternative(self, ref):
 
 	def getBouquetList(self):
 		bouquets = []
-		if self.isSubservices():
- 			bouquets.append((self.getServiceName(self.subservicesBouquet), self.subservicesBouquet))
 		serviceHandler = eServiceCenter.getInstance()
```
jbleyel, line 637: Why do you have removed this?

jbleyel:
This change will break many other things. You cannot add all services to the results of getBouquetList because "all services" is not a bouquet.

PR #3310, opened 2024-05-04
l10n wxyz

wedebe:
This change refreshes English text strings into more readable, concise and user-friendly language for a better overall experience.

Focus here is on files beginning *w, x, y, or z*, but other files using modified strings will also have been updated.

### DIFF
--- a/lib/python/Screens/Wizard.py
+++ b/lib/python/Screens/Wizard.py
@@ -54,7 +54,7 @@ class Wizard(Screen, HelpableScreen):
 		self.showStepSlider = showStepSlider
 		self.showList = showList
 		self.showConfig = showConfig
-		self.setTitle(_("Welcome Wizard"))
+		self.setTitle(_("Setup Wizard"))
 		self.wizard = {}
 		parser = make_parser()
 		wizardHandler = self.parseWizard(self.wizard)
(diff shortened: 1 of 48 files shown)


IanSav:
I don't think these changes are consistent with the rest of Enigma.  "Changing "Welcome" to "Setup" creates confusion with the whole "Setup" configuration system.  They are different things.

Also, changing "Proceed" to "Move" is inconsistent with the usage of "Move" to indicate cursor navigation within the screen.  Similarly, Changing "Move" to "Go to" is inconsistent with the current Enigma navigation help.

The "Title" widget is available to any skins that want to display it so setTitle() is neecded.

Please be aware that UserInterfacePositioner.py is gone in a commit that is about to be merged.

wedebe:
> I don't think these changes are consistent with the rest of Enigma. "Changing "Welcome" to "Setup" creates confusion with the whole "Setup" configuration system. They are different things.
> 
> Also, changing "Proceed" to "Move" is inconsistent with the usage of "Move" to indicate cursor navigation within the screen. Similarly, Changing "Move" to "Go to" is inconsistent with the current Enigma navigation help.
> 
> The "Title" widget is available to any skins that want to display it so setTitle() is neecded.
> 
> Please be aware that UserInterfacePositioner.py is gone in a commit that is about to be merged.

This is the beginning of textual improvements I'd like to made across the openATV codebase, with the intention of providing the best user experience, but with an eventual goal of having every visible piece of text become familiar and consistent across every screen/menu/label/button etc. I know there's been work done previously, but there appears to be a lot of literal translations done with the best of intentions which have been... *fudged* into not-so-natural English (stuff a native-tongue English person wouldn't really say) and tweaked slightly.

Indeed the very reason I started reverse alphabetically was so there'd be a minimum amount of files affected so as to attract feedback :)

RE Welcome/Setup, the reasoning behind this is...
- anything that involves initial configuration/initialisation should be called *setup* (eg. MacOS Setup)
- anything that involves steps should be called a *wizard* (eg. File Transfer Wizard) 
- anything that involves reconfigurable stuff should be called *settings* (eg. Menu | Settings | System Settings)

BTW, I've previously done a heapload of work on translation which was in use for ages only for it to be wiped when en.po was no longer rendered 😭
https://github.com/openatv/enigma2/commits/7c8c88df5775cde7ec516e2765d5286f54a7828a/po/en.po?author=wedebe

jbleyel:
Hi @wedebe , you miss one important thing.
The English text is also used for translations.
That's why we should always use text which is cristal clear as possible to be translated.

By the way .. many "Setup" are translated to "Einstellungen" in German and "Settings" is also "Einstellungen".
This is only one example and one Language!!!

IanSav:
@wedebe, I think some of my issue with the text changes is that they are coming out of order and out of context.  If you want to change navigation text then all the navigation text should be changed at the same time.  That is, don't change text alphabetically but by functionality.  For example, if you want to change navigation help text then change all the navigation help text in all the code and places where these text messages are used.  That way, as a user continues to use Enigma all the navigation screens will change at once and remain consistent across the UI.

It would also help if you can document a plan of what types of text you will be changing and how you propose to change that text,  This plan can be discussed and updated, if required, so that you don't create changes that may conflict with other intentions.  For example, again using navigation as the example, provide the list of all 10 common navigation options and how you plan to change them across the whole UI.  Where those navigation buttons have different meanings then how does the text change.  I am thinking about times where LEFT and RIGHT move left and right along a list of options or when LEFT and RIGHT moves between left and right panels on the screen.  You wanted to use "Go to" for navigation instead of "Move" then you need to change it uniformly.  In your plan you would indicate that Navigation should use the term "Go to".  Similarly "Move" is used to ..., etc.

May I suggest an alternative strategy.  Why don't you take a text rich area of the code and change it as a single step.  My suggestion would be to go through setup.xml and make all the options and language concise, correct and informative.  Where a description for an item like "Set option" has a description like "This sets option." then scrap that description and document what the option actually does and what the various options may mean or do.  This would be a high value change with minimal disruption.  Don't try to correct the ".po" files as @jbleyel has some scripts that does these changes in a way that causes minimum issues for translators.  You should collect all changes in a text file with old / new text like this:
`"OLDTEXT"="NEWTEXT"`
Then send the file to @jbleyel for processing in his script.

All text manipulation is monitored and managed by @jbleyel and there are some important management aspects that must be followed.

PR #3338, opened 2024-05-28
[Screen]

jbleyel:
* add ignoreWidgets attribute handling

This allows you to ignore non used widgets.

### DIFF
--- a/lib/python/Screens/Screen.py
+++ b/lib/python/Screens/Screen.py
@@ -272,12 +272,12 @@ class Screen(dict):
 		zPosition = 0
 		for (key, value) in self.skinAttributes:
 			match key:
+				case "ignoreWidgets":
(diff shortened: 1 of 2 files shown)


lib/python/Screens/Screen.py
```diff
@@ -275,6 +276,8 @@ def applySkin(self):
 					resolution = tuple([int(x.strip()) for x in value.split(",")])
 				case "zPosition":
 					zPosition = int(value)
+				case "ignoreWidgets":
+					self.ignoreWidgets = [x.strip() for x in value.split(",")]
```
IanSav, line 280: It would be better if the cases were sorted alphabetically.

IanSav (approved):
The changes are fine.  I have only one comment/suggestion to make the case list in Screen.py alphabetical.  This is obviously not a code issue just style feedback.

PR #3363, opened 2024-06-19
[MultiBoot]

jbleyel:
- Use f-string
- Fix struct error
- Fix bootloop if boot partition is not on device sda ( #3361 )

### DIFF
--- a/lib/python/Tools/MultiBoot.py
+++ b/lib/python/Tools/MultiBoot.py
@@ -199,11 +199,12 @@ class MultiBootClass():
 								bootSlots[slotCode]["ubi"] = True
 							if "UUID=" in line:
 								bootSlots[slotCode]["uuid"] = True
+							bootDevice = [x for x in ("sda", "sdb", "sdc", "sdd") if x in line]
 							if "rootsubdir" in line:
 								bootSlots[slotCode]["kernel"] = self.getParam(line, "kernel")
 								bootSlots[slotCode]["rootsubdir"] = self.getParam(line, "rootsubdir")
-							elif [x for x in ("sda", "sdb", "sdc", "sdd") if x in line]:
-								device = [x for x in ("sda", "sdb", "sdc", "sdd") if x in line][0]
+							elif bootDevice:
+								device = bootDevice[0]
 								saveKernel(bootSlots, slotCode, f"/dev/{device}{line.split(device, 1)[1].split(" ", 1)[0]}")
 							else:
 								parts = device.split("p")
--- a/lib/python/Tools/MultiBoot.py
+++ b/lib/python/Tools/MultiBoot.py
@@ -3,7 +3,8 @@ from glob import glob
 from hashlib import md5
 from os import mkdir, rename, rmdir, stat
 from os.path import basename, exists, isdir, isfile, ismount, join as pathjoin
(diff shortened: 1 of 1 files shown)


lib/python/Tools/MultiBoot.py
```diff
@@ -202,16 +203,17 @@ def saveKernel(bootSlots, slotCode, kernel):
 							if "rootsubdir" in line:
 								bootSlots[slotCode]["kernel"] = self.getParam(line, "kernel")
 								bootSlots[slotCode]["rootsubdir"] = self.getParam(line, "rootsubdir")
-							elif "sda" in line:
-								saveKernel(bootSlots, slotCode, "/dev/sda%s" % line.split("sda", 1)[1].split(" ", 1)[0])
+							elif match("sd[abcd]", line):
```
IanSav, line 206: Rather than using a regular expression, would a line.startswith((sda, sdb, sdc, sdd)) be faster?

IanSav (approved):
As commented within the code, I don't think a regular expression import and processing is needed.

lib/python/Tools/MultiBoot.py
luisteraar: Thanks for looking into it

in the future there is also a small fix in file /etc/udev/mount-helper.sh
Blacklist boot device
# detected multiboot sdcard
                        if [ $stbcheck == "viper4k" ] || [ $stbcheck == "sf8008" ] || [ $stbcheck == "sf8008m" ] || [ $stbcheck == "ustym4kpro" ] || [ $stbcheck == "beyonwizv2" ] || [ $stbcheck == "gbmv200" ]; then
                                DEVCHECK=`expr substr $MDEV 1 3`
                                if [ "${DEVCHECK}" == "sda" ] ; then
                                        DEVSIZE=`cat /sys/block/sda/sda1/size`
                                        if [ $DEVSIZE -lt "32769" ]; then
                                                BLACKLISTED=`echo ${BLACKLISTED} sda`
                                        fi
                                fi

lib/python/Tools/MultiBoot.py
```diff
@@ -203,8 +202,8 @@ def saveKernel(bootSlots, slotCode, kernel):
 							if "rootsubdir" in line:
 								bootSlots[slotCode]["kernel"] = self.getParam(line, "kernel")
 								bootSlots[slotCode]["rootsubdir"] = self.getParam(line, "rootsubdir")
-							elif match("sd[abcd]", line):
-								device = match("sd[abcd]", line).group(0)
+							elif [x for x in ("sda", "sdb", "sdc", "sdd") if x in line]:
+								device = [x for x in ("sda", "sdb", "sdc", "sdd") if x in line][0]
```
IanSav, line 206: This can be optimized.

bootDevice = [x for x in ("sda", "sdb", "sdc", "sdd") if x in line]
...
elif bootDevice:
    device = bootDevice[0]

IanSav (approved):
There is an optimization for this latest update.  A code fragment suggested in Skype.

lib/python/Tools/MultiBoot.py
luisteraar: I'm not sure but can we get rid of the restriction for sf4008/sf8008 etc and for boxes with 
machine_features emmc

Issue #3383, opened 2024-07-14
Crash durung settings backup


ian10951:
If during a settings backup you press the green button to display the tv picture whilst the backup continues in the background, but then forget you are doing a backup and press the OK button the box crashes.

Crashlog attached.
[20240714-084926-enigma2-crash.log](https://github.com/user-attachments/files/16216004/20240714-084926-enigma2-crash.log)

IanSav:
Fixed in #3384

PR #3384, opened 2024-07-14
[Console.py] Forrect crash in #3383

IanSav:
This change corrects the location of the initialization of the self.baseTitle variable to resolve the reported crash in Issue #3383.


IanSav:
This was a duplicated fix.

PR #3387, opened 2024-07-21
Update uk.po

badsystem:


jbleyel:
we are using weblate only for translation

badsystem:
it is not possible to add a line for translation there

jbleyel:
That’s not working if this line is not in the pot file. Where is this text?

badsystem:
The text itself is in lib/python/Tools/ISO639.py
Outputs to AudioSelect

jbleyel:
The values from ISO639 are not prepared for translation.
That means your change will have no effect.

PR #3435, opened 2024-09-30
[Converter/ServiceInfo] Fixed HasHBBTV

m4dhouse:


m4dhouse:
> iPlayableService.evHBBTVInfo should be enough as trigger

It does not update the information, it always remains enabled.

pippomus:
Madhouse 
number one

PR #3466, opened 2024-12-08
[OpenTV]

jbleyel:
* add genre/rating support

Cherry Pick : https://github.com/OpenViX/enigma2/commit/2a92646cacf5922dc150f2688b48e644f45dc3f1


lib/python/Components/UsageConfig.py
AbuBaniaz, line 1: I think using Great Britain is not correct. This applies to Republic of Ireland too.

jbleyel: The code is from Vix.

PR #3472, opened 2024-12-14
[Components/About] Fixed "Last Flash" information

m4dhouse:


jbleyel:
I prefer a different file or folder and nothing inside of /etc.

PR #3505, opened 2025-01-21
[Information]

m4dhouse:
Fixed crash if capabilities variable is None

### DIFF
--- a/lib/python/Screens/Information.py
+++ b/lib/python/Screens/Information.py
@@ -2240,12 +2240,13 @@ class TunerInformation(InformationBase):
 			return f"{format_string(format, value, grouping=True)} {units}" if units else format_string(format, value, grouping=True)
 
 		def extractModes(data, mode):
-			mode = f"{mode} "
-			length = len(mode)
 			values = []
-			for item in data.split(","):
-				if item.startswith(mode):
-					values.append(item[length:].capitalize())
+			if data:
+				mode = f"{mode} "
+				length = len(mode)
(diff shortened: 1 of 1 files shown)


lib/python/Screens/Information.py
```diff
@@ -2240,6 +2240,8 @@ def formatNumber(number):
 			return f"{format_string(format, value, grouping=True)} {units}" if units else format_string(format, value, grouping=True)
 
 		def extractModes(data, mode):
+			if not data:
+				return
```
jbleyel, line 2244: This is OK.

lib/python/Screens/Information.py
```diff
@@ -2249,12 +2251,14 @@ def extractModes(data, mode):
...
-				addAuto = False
-			values = [str(x) for x in sorted([int(x) for x in values])]
+			if not values:
+				return
+			addAuto = "Auto" in values
+			values = [x for x in values if x != "Auto"]
+			try:
+				values = [str(x) for x in sorted(map(int, values))]
+			except ValueError:
+				pass
```
jbleyel, line 2261: Why do you have changed that?

jbleyel:
Sorry, I have changed the file and you need to update your pull.

lib/python/Screens/Information.py
m4dhouse: My mistake, they are just numerical values, that's not necessary, you can close the PR

PR #3511, opened 2025-01-28
[Directories]

jbleyel:
* use different skinResolveList for xml files to support old skins


IanSav (approved):
This is a simple and logical fix to address legacy skin feature from the past and into the future.

PR #3512, opened 2025-01-29
[International]

jbleyel:
* improve boot time
* remove blocking popen


IanSav (approved):
Can a note be added to the build system to warn people that if the locales change then International.py needs to be updated?

Issue #3504, opened 2025-01-21
MuteDialog on Infobar


GhostofGeeeee:
Steps:
- do mute
- press OK(Infobar and muteDialog is shown)
- wait
	- Infobar closes in (config.usage.infobar_timeout) seconds
	- muteDialog closes in (config.volumeControl.hideTimer) seconds

I know this are different settings but, I think, in this case the muteDialog should have the same timeout.

urie57:
Personally if I want to mute I want it instantly not have a timeout or have option of instant if time out is implemented

jbleyel:
> Personally if I want to mute I want it instantly not have a timeout or have option of instant if time out is implemented

Do you really understand the issue?
This is about the mute pop up window close timeout.

urie57:
> > Personally if I want to mute I want it instantly not have a timeout or have option of instant if time out is implemented
> 
> Do you really understand the issue? This is about the mute pop up window close timeout.

Yes I understand it now , but if I press mute then press ok to display infobar they are at slightly different times it is two different ke ypresses how can they ever be the same time.

IanSav:
The different screen objects are reacting to completely different triggers.  Linking their timeouts is unwise.  For example, you press OK to bring up the InfoBar and a few seconds later you press MUTE.  If they are linked then you might never see the MUTE icon as its timer will be overridden for the InfoBar timer.

Even if we ignore the user interface issues that this would create then we would also create a coding issue of trying to synchronise the UIs of two completely unrelated objects.

I have discussed this with another developer and we agree that this will not a feature that will benefit the usability of the system.

GhostofGeeeee:
I accept, maybe feature is done.
But this [design](https://github.com/openatv/enigma2/blame/69eb9df67e2e88760b1ee344c64afc4455b6574a/lib/python/Screens/InfoBarGenerics.py#L830) is rudimental and could be a bug, please reopen more later.

GhostofGeeeee:
More is something like:
Steps:
- do mute
- press OK(Infobar and muteDialog is shown)
- open secondInfobar or Channelselection..., the muteDialog don't hide.

urie57:
I do not have that problem the mute dialog disappears after time set in Audio > Audio Settings > Volume/Mute display timer

urie57:
> True but during this period, it's on top.

To be honest I never even noticed it as I had HDMI-CEC enabled and my volume bar is dispalyed on right hand side was looking over that way

[...]

PR #3515, opened 2025-02-02
Link/connect Infobar and muteDialog

GhostofGeeeee:
Issue #3504 Infobar opens the muteDialog, but couldn't handle it.
Now Infobar open and close it together.
The muteDialog will not longer coverup other Screens while its open, this is only desired for a volume action.
Fix a Bug that an open Infobar cancels the volumeMuteLong action.

### DIFF
--- a/lib/python/Screens/InfoBarGenerics.py
+++ b/lib/python/Screens/InfoBarGenerics.py
@@ -827,7 +827,6 @@ class InfoBarShowHide(InfoBarScreenSaver):
 		for x in self.onShowHideNotifiers:
 			x(True)
 		self.startHideTimer()
-		VolumeControl.instance and VolumeControl.instance.showMute()
 
 	def doDimming(self):
 		if config.usage.show_infobar_do_dimming.value:
--- a/lib/python/Components/VolumeControl.py
+++ b/lib/python/Components/VolumeControl.py
@@ -97,11 +97,6 @@ class VolumeControl:
 		config.volumeControl.volume.setValue(self.dvbVolumeControl.getVolume())
 		config.volumeControl.save()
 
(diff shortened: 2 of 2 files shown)


GhostofGeeeee:
Is there any feedback?
Close?
Is the "my preference" commit a problem? I'll revert it!

jbleyel:
> Is there any feedback? Close? Is the ["my preference"](https://github.com/openatv/enigma2/pull/3515/commits/811cca5353540817e5099a19b20f86489e4f0bef) commit a problem? I'll revert it!

The Infobar should NOT trigger any show/hide actions in the Mute / Volume dialog.
These are complete independent things and should be not mixed in any form.

lib/python/Screens/InfoBarGenerics.py
```diff
@@ -827,7 +827,6 @@ def __onShow(self):
 		for x in self.onShowHideNotifiers:
 			x(True)
 		self.startHideTimer()
-		VolumeControl.instance and VolumeControl.instance.showMute()
```
jbleyel, line 830: This is OK, the Infobar should not show the Mute.

GhostofGeeeee: LOL, OK, that's easy! I agree.
[def showMute(self):  # This method is only called by InfoBarGenerics.py:](https://github.com/openatv/enigma2/blob/f6cb7dd0cb436e1259dc094ffb2b7ea94334359d/lib/python/Components/VolumeControl.py#L100)
I'm not sure about 3rd party but don't forget!!!

IanSav (approved):
Given that no-one has provided any reason why pressing OK should display the MUTE status (and not the volume status) then I see no reason why the two unrelated actions can be separated.

PR #3517, opened 2025-02-06
[FileCommander] Fix's for broken symlinks.

GhostofGeeeee:
Issue:
- create symlink
- delete target
- try to del, move, copy, info, menu, etc. for symlink.

### DIFF
--- a/lib/python/Plugins/Extensions/FileCommander/plugin.py
+++ b/lib/python/Plugins/Extensions/FileCommander/plugin.py
@@ -936,8 +936,8 @@ class FileCommander(Screen, NumericalTextInput, StatInfo):
 	def keyHashes(self, path=None):
 		if path is None:
 			path = self.sourceColumn.getPath()
-		if isfile(path):
-			if stat(path).st_size < HASH_CHECK_SIZE:
+		if isfile(path) or islink(path):
+			if lstat(path).st_size < HASH_CHECK_SIZE:
 				import hashlib
 				data = {}
 				data["Screen"] = "FileCommanderHashes"
@@ -2879,7 +2879,7 @@ class FileDeleteTask(Job):
 		count = len(srcPaths)
 		for index, srcPath in enumerate(srcPaths):
 			taskName = _("Directory/File %d of %d" % (index + 1, count))
-			if isfile(srcPath):
+			if isfile(srcPath) or islink(srcPath):
 				FileTransferTask(self, taskName, srcPath, dirname(normpath(srcPath)), FileTransferTask.JOB_DELETE)
 			else:
 				FileTransferTask(self, taskName, srcPath, dirname(normpath(srcPath)), FileTransferTask.JOB_DELETE_TREE)
@@ -2910,7 +2910,7 @@ class FileTransferTask(Task):
 
 	def __init__(self, job, taskName, srcPath, dstPath, jobType):
 		Task.__init__(self, job, taskName)
-		if exists(srcPath) and exists(dstPath):
+		if lexists(srcPath) and exists(dstPath):
 			self.srcPath = srcPath
 			self.dstPath = dstPath
 			target = join(dstPath, "") if isdir(srcPath) else join(dstPath, basename(normpath(srcPath)))
@@ -2949,7 +2949,7 @@ class FileTransferTask(Task):
 		self.progressTimer.start(self.updateTime, True)
 
 	def prepare(self):
-		self.srcSize = float(self.dirSize(self.srcPath) if isdir(self.srcPath) else getsize(self.srcPath))
+		self.srcSize = float(self.dirSize(self.srcPath) if isdir(self.srcPath) else lstat(self.srcPath).st_size)
 		self.updateTime = max(1000, int(self.srcSize * 0.000001 * 0.5))  # Based on 20Mb/s transfer rate.
 		self.progressTimer.start(self.updateTime, True)
 


jbleyel:
Please provide more info:

What's the exact use case of this change?
Please provide an example.

GhostofGeeeee:
Issue:
No actions possible for a broken symlink(create symlink, delete target)

Steps to reproduce:
- create symlink
- delete target
- try to del, move, copy, info, menu, etc. for symlink. --> displayStatus(_("The directory/file no longer exists."))

It's not a user request. I had a broken symlink but i couldn't delete it.

Now it should be possible to delete, move, rename, copy it. It's possible to open Menu and do Hash/Checksum action.
I won't create a exception for "Change execute permissions (755/644)"

Thank you!

jbleyel:
@IanSav will take over this.

IanSav (commented):
This code will not work as expected.  While it supresses the error it may have unexpected effects.

I have corrected the original issue in another pull request.

jbleyel:
https://github.com/openatv/enigma2/pull/3523

GhostofGeeeee:
Alot of f-Strings again, I'm searching the changes for this issue, the icon is good but you only pick one change of mine and missed the rest:
	def checkStillExists(self, path):
		if path and not lexists(path):
I think my code is better with less unexpected effects.

GhostofGeeeee:
I can press delete and confirm but link is still there.
+
> do Hash/Checksum action

IanSav:
There is a lot of work to do and very few people to do it.  When possible, any code that is touched is also brought up to current coding standards. f-string updates have a performance enhancement benefit!

I have no idea what "do Hash/Checksum action" means.  If this is meant to be a bug report I suggest you format it as a new bug report with all the details required to allow the issue to be replicated together with what actions are expected.

GhostofGeeeee:
I spent some work to do this fix.
> This code will not work as expected. While it supresses the error it may have unexpected effects.

"have unexpected effects" is not clear to me.
No problem, if you do it better but it's not the case.

But "This code will not work as expected." looks that you had a look on it.

GhostofGeeeee:
I spent some work to do this fix.
> This code will not work as expected. While it supresses the error it may have unexpected effects.

"have unexpected effects" is not clear to me.
No problem, if you do it better but it's not the case.

But "This code will not work as expected." looks that you had a look on it.

"Fix "The directory/file no longer exists." for broken symlinks." was the entry issue but followed by more, I think you know this.

**I need to say I only do draft's because I learned you can do it better.**
My documentation was bad/ not finished.
But I can't explain these small fixes with a big documentaion.

I found this Issue and I solved all I found.
Now your help was welcome to finish it.

lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -2879,7 +2879,7 @@ def __init__(self, srcPaths, title):
 		count = len(srcPaths)
 		for index, srcPath in enumerate(srcPaths):
 			taskName = _("Directory/File %d of %d" % (index + 1, count))
-			if isfile(srcPath):
+			if isfile(srcPath) or islink(srcPath):
```
jbleyel, line 2882: This is wrong .. try islink on a folder link.

lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -2910,7 +2910,7 @@ class FileTransferTask(Task):
 
 	def __init__(self, job, taskName, srcPath, dstPath, jobType):
 		Task.__init__(self, job, taskName)
-		if exists(srcPath) and exists(dstPath):
+		if lexists(srcPath) and exists(dstPath):
```
jbleyel, line 2913: Why do you do this?

lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -2949,7 +2949,7 @@ def progressUpdate(self):
 		self.progressTimer.start(self.updateTime, True)
 
 	def prepare(self):
-		self.srcSize = float(self.dirSize(self.srcPath) if isdir(self.srcPath) else getsize(self.srcPath))
+		self.srcSize = float(self.dirSize(self.srcPath) if isdir(self.srcPath) else lstat(self.srcPath).st_size)
```
jbleyel, line 2952: Are you sure? We want the size of the real file not the size of the link.

lib/python/Plugins/Extensions/FileCommander/plugin.py
GhostofGeeeee: It returns True!?

lib/python/Plugins/Extensions/FileCommander/plugin.py
GhostofGeeeee: Why? Why you don't do this!!!!!!

lib/python/Plugins/Extensions/FileCommander/plugin.py
GhostofGeeeee: This is a idea but it will not follow the link!

lib/python/Plugins/Extensions/FileCommander/plugin.py
jbleyel: So in case of a folder you only want to delete the link.
That's probably ok.

lib/python/Plugins/Extensions/FileCommander/plugin.py
jbleyel: The question is more like, what should happen if you want to copy/move/delete a link?

lib/python/Plugins/Extensions/FileCommander/plugin.py
jbleyel: Have you tested the progress?
You may need to add debug log to see the progress.

lib/python/Plugins/Extensions/FileCommander/plugin.py
```diff
@@ -936,8 +936,8 @@ def keyGoBottom(self):
 	def keyHashes(self, path=None):
 		if path is None:
 			path = self.sourceColumn.getPath()
-		if isfile(path):
-			if stat(path).st_size < HASH_CHECK_SIZE:
+		if isfile(path) or islink(path):
+			if lstat(path).st_size < HASH_CHECK_SIZE:
```
jbleyel, line 940: So you want to get the has from the target and not from the link?
Then you need to do more things.

lib/python/Plugins/Extensions/FileCommander/plugin.py
GhostofGeeeee: On my test I expected the size of the link and with my change I got it.

lib/python/Plugins/Extensions/FileCommander/plugin.py
jbleyel, line 940: This is wrong if you ask me.

jbleyel: islink is true for a directory link and this is wrong.
Creating a hash from a link makes no sense.

lib/python/Plugins/Extensions/FileCommander/plugin.py
jbleyel: Then we may need to fix getsize.

jbleyel: But it seams that lstat and getsize gives you the same result on a file.
So your change looks ok.

lib/python/Plugins/Extensions/FileCommander/plugin.py
jbleyel: We can think about to create the hash from the link target if the link points to a file.

jbleyel:
Is this pull ready or do you want to add more and make more tests?

GhostofGeeeee:
I can't say that a problem is impossible but i can't find a problem!

PR #3520, opened 2025-02-10
Add translation for "Playback Time:" on EventViewSimple Screen

GhostofGeeeee:
https://github.com/openatv/MetrixHD/blob/ad2959857023e5bf5ef5c80d7c88e4cbd26b2085/usr/share/enigma2/MetrixHD/skinfiles/skin_openatv.xml#L2314


jbleyel:
The pot file will be automatically generated.
So your change will be reverted.

PR #3525, opened 2025-02-13
[InfoBarPiP] showPiP() code improvement

GhostofGeeeee:
- hide swapPiP extension if unavailable
- better PiP support for EMC, IPTV plugins

### DIFF
--- a/lib/python/Screens/InfoBar.py
+++ b/lib/python/Screens/InfoBar.py
@@ -805,38 +805,6 @@ class MoviePlayer(InfoBarAspectSelection, InfoBarSimpleEventView, InfoBarBase, I
 		else:
 			InfoBarSeek.seekBack(self)
 
-	def showPiP(self):
-		slist = self.servicelist
-		if self.session.pipshown:
-			if slist and slist.dopipzap:
-				slist.togglePipzap()
-			if self.session.pipshown:
-				del self.session.pip
-				self.session.pipshown = False
-		else:
-			service = self.session.nav.getCurrentService()
-			info = service and service.info()
-			xres = str(info.getInfo(enigma.iServiceInformation.sVideoWidth))
-			if int(xres) <= 720 or BoxInfo.getItem("model") != 'blackbox7405':
-				from Screens.PictureInPicture import PictureInPicture
-				self.session.pip = self.session.instantiateDialog(PictureInPicture)
-				self.session.pip.show()
-				if self.session.pip.playService(slist.getCurrentSelection()):
-					self.session.pipshown = True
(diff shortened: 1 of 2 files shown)


jbleyel:
Have you tested this code?
It's not working.
You may should do more tests.

jbleyel:
You can't use isSeekable instead of isStandardInfoBar.

GhostofGeeeee:
Yes I did test it. Please give me a hint!

I'm not sure for what's the case for `newservice = "self.session.nav.getCurrentlyPlayingServiceReference()"`
Is this a current wanted behavior, to have a CurrentlyPlayingServiceReference() thats not self.servicelist.servicelist.getCurrent()
I think it could be a hack to change the service without servicelist.

Live TV is not Seekable and we can have it easy on pip, is it a movie or stream we should not open it on pip(movie don't work)(stream buffer can be reused?) But if you say no, good to know and I can remove my thoughts.

jbleyel:
> Yes I did test it. Please give me a hint!
> 
> I'm not sure for what's the case for `newservice = "self.session.nav.getCurrentlyPlayingServiceReference()"` Is this a current wanted behavior, to have a CurrentlyPlayingServiceReference() thats not self.servicelist.servicelist.getCurrent() I think it could be a hack to change the service without servicelist.
> 
> Live TV is not Seekable and we can have it easy on pip, is it a movie or stream we should not open it on pip(movie don't work)(stream buffer can be reused?) But if you say no, good to know and I can remove my thoughts.

self.servicelist.servicelist.getCurrent will not work for MoviePlayer.

jbleyel:
> > Live TV is not Seekable and we can have it easy on pip, is it a movie or stream we should not open it on pip(movie don't work)(stream buffer can be reused?) But if you say no, good to know and I can remove my thoughts.
> 

You miss one point.
InfoBar Player is also used for IPTV VOD and that's Seekable.

jbleyel:
You need to test all possible service types and not only Live TV.

GhostofGeeeee:
> InfoBar Player is also used for IPTV VOD and that's Seekable.

Ah, good thank you!
~~Seekable~~

> You need to test all possible service types and not only Live TV.

Sorry my use is limited. 

>It's not working.

What does it mean?

GhostofGeeeee:
With your feedback I will change it. Both?:
```
				if isStandardInfoBar(self):
					newservice = self.lastPiPService or self.session.nav.getCurrentlyPlayingServiceReference() or self.servicelist.getCurrentSelection()
				else:
					newservice = self.lastPiPService or self.servicelist.getCurrentSelection()
```

PR #3545, opened 2025-02-25
[VolumeControl.py] Correct a comment

IanSav:


jbleyel:
This should be a bigger commit with more changes.
A commit with one line especially only comments is not really needed.
If you ask me.

IanSav:
This module was just updated.  This is simply a correction that was missed in the recent commit.

More changes are not currently planned, unless the legacy methods can now be removed.

PR #3546, opened 2025-03-02
[Timeshift] checkTimeshiftRunningCallback

GhostofGeeeee:
- the callback trace was broken but needed.

https://www.opena.tv/viewtopic.php?p=581160#p581160 https://github.com/openatv/enigma2/blob/a4861c6ab6faf97bfb001b9304d8f8b80ad41597/lib/python/Screens/ChannelSelection.py#L2603

### DIFF
--- a/lib/python/Components/Timeshift.py
+++ b/lib/python/Components/Timeshift.py
@@ -512,7 +512,7 @@ class InfoBarTimeshift:
 				self.save_current_timeshift = False
 			elif answer == "no":
 				pass
-			InfoBarTimeshift.saveTimeshiftActions(self, answer, returnFunction)
+		InfoBarTimeshift.saveTimeshiftActions(self, answer, returnFunction)
 
 	def eraseTimeshiftFile(self):
 		for filename in listdir(config.timeshift.path.value):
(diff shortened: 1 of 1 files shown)


lib/python/Components/Timeshift.py
```diff
@@ -657,7 +657,7 @@ def saveTimeshiftActions(self, action=None, returnFunction=None):
 		elif action == "no":
 			pass
 		# Get rid of old time shift file before E2 truncates its filesize
-		if returnFunction is not None and action != "no":
+		if returnFunction is not None and action and action != "no":
```
jbleyel, line 660: Are you sure becasue your change may have negative effects:

```
>>> action1 = None
>>> action2 = ""
>>> action3 = "no"
>>> action1 and action1 != "no"
>>> action2 and action2 != "no"
''
>>> action3 and action3 != "no"
False
>>> action1 != "no"
True
>>> action2 != "no"
True
>>> action3 != "no"
False
```

GhostofGeeeee: (Sorry, I'm late, Discord Github Bot don't read these comments.)
This should prevent negative effects.

GhostofGeeeee: This [check](https://github.com/openatv/enigma2/pull/3546/files#diff-2faa1d730d81bad48ee05e47e5c70ee45b1fa0b0b8e0e94dd88a1b18fdd43d80R508) is just skiped and moves to this new location as action.

jbleyel: Yes, but if answer is None then you don't call erase.
Is that correct?

GhostofGeeeee: It wasn't called before, I think so!

PR #3560, opened 2025-04-20
Update MultiBoot.py

dos-expert:
corrected VTi version info

### DIFF
--- a/lib/python/Tools/MultiBoot.py
+++ b/lib/python/Tools/MultiBoot.py
@@ -440,7 +440,7 @@ class MultiBootClass():
 				compileDate = f"{compileDate[0:4]}-{compileDate[4:6]}-{compileDate[6:8]}"
 				version = str(info.get("version"))
 				if "." not in version and "-" not in version and version.isdigit():
-					version = f"{int(version[0:2])}.{int(version[3:5])}"
+					version = f"{int(version[0:2])}.{version[2:3]}.{version[3:5]}"
 				self.imageList[self.slotCode]["detection"] = "Found an image version file"
 				creator = info.get("creator")
 				if creator is not None:


dos-expert:
before: VTi 15.2
after: VTi 15.0.02 (which is correct)

IanSav:
Can you please post the content of a sample VTi "/etc/image-version" file.  Is this file typical of all VTi versions of this file?  How many styles has the version string ever had on VTi?

dos-expert:
All I can do is give you the content my VTi "/etc/image-version" file here (Vu+ Ultimo4K):

> version=015002202212231303
> creator=VTi <info@vuplus-support.org>
> url=http://www.vuplus-support.org/

Also for your info: the previous code is also from me. I sent it directly to Captain at that time.
So this is a correction of my own code.
Additional info: the bytes beginning on offset 6 ('202212231303') represent the release date (2022/12/23 13:03)
see for example [here](https://www.vuplus-support.org/wbb4/vtisoftware/?dir=VTi%2015.0.X/VTi%2015.0.02/Vu%2B%20Ultimo%204K)

IanSav:
So you would like the version number to be displayed as "15.2 (2022/12/23 13:03)"?

IanSav:
The fix has been incorporated with other update via https://github.com/openatv/enigma2/pull/3566 .

dos-expert:
Sorry that was not my code and still didn't work as expected.

Still: VTi 15.2
Why? 
content of "/etc/image-version":
version=015002202212231303
content of "version" after "version = str(info.get("version"))":
'15002202212231303'
so len(version) = 17 (trailing zero vanishing)
so your line:
version = f"{int(version[0:3])}.{version[3:4]}.{version[4:6]}" if len(version) == 18 else f"{int(version[0:2])}.{int(version[3:5])}"
will return the second part which is "15.2"

better distinguish three cases:
1. major version < 10:
  len(version) == 16:
    version = f"{version[0:1]}.{version[1:2]}.{version[2:4]}"
2. major version <100:
  len(version) == 17:
    version = f"{version[0:2]}.{version[2:3]}.{version[3:5]}"
3. major version >= 100:
  len(version) == 18:
    version = f"{version[0:3]}.{version[3:4]}.{version[4:6]}"

thanks for you patience

jbleyel:
Please try this:

`version = f"{int(version[0:3])}.{version[3:4]}.{version[4:6]}" if len(version) == 17 else f"{int(version[0:2])}.{int(version[3:5])}"
`

IanSav:
The issue is not where you highlighted.  The issue is that "version=015002202212231303" is all numeric so it is treated as a number.  If it was meant to be a string it should have been "version="015002202212231303"".  This would preserve all the characters.

If the VTi version numbers were consistent then this would not be an issue to identify.  The fix suggested by @jbleyel should address the issue as there is only one version of VTi that is likely to be accessed.

jbleyel:
Already fixed -> https://github.com/openatv/enigma2/commit/6502d5f8483ceaf8a85eafeea09390ae737a8065

dos-expert:
The fix makes it even worse.
Before: VTi 15.2
Now: VTi 150.0.22

Goal: VTi 15.0.02

Please read carefully my last comment.
I also vote for the suggestion of @IanSav, then you don't have to distinguish the cases I mentioned above.

dos-expert:
here my new suggestion:
('mvl' stands for major version length (the first part of the version) and should be 1, 2 or 3)
> mvl = len(version)-15
> version = f"{int(version[0:mvl])}.{version[mvl:mvl+1]}.{version[mvl+1:mvl+3]}"

Example:
version="15002202212231303"
mvl=2
result: "15.0.02"

dos-expert:
Supplement: you can remove the int() function in my suggestion above

jbleyel:
I think we should stop this here because we have many other and real issues.

PR #3571, opened 2025-05-03
[OScamInfo] * fix unbound variable

WXbet:
* see https://www.opena.tv/viewtopic.php?p=584037#p584037


jbleyel:
We make cherry pick from master.

PR #3572, opened 2025-05-06
[About] * add Samba and UPX version

WXbet:


WXbet:
### Samba version information
image

### File compression
image
image

"File compression" needs to be translated.

PR #3593, opened 2025-06-09
[keymap.xml] Add keyboard button definitions for volume control

IanSav:


IanSav:
I neglected to mention the button definitions.  The keyboard keypad buttons are defined as:

- "-" -> Volume up
- "+" -> Volume down
- "*" -> Toggle mute
- "*" (Long) -> Activate mute and keep the MUTE icon on the screen.

PR #3609, opened 2025-07-22
[InfoBarGenerics.py] Replace "Seekbar" screen with "SeekBar" screen

IanSav:
This new screen implements a whole new mechanism for seeking within media files.  Users access the new screen in the same way that they accessed the old "Seekbar" screen as the new screen is intended as a complete replacement.

The new screen is significantly more configurable and has an added HELP screen to show what actions are currently available for the settings and modes selected.

In the Playback settings some of the new options that were recently added now become active.

Firstly, the actions of the ARROW buttons is now configurable.  The "Arrow skip mode" setting has options of "Traditional" where the ARROW buttons act as they have always worked, "Symmetrical skips" where the sensibility percentage skips are applied, and "Defined skips where each ARROW button can be assigned to skip any number of seconds in either direction.

The ARROW button sensibility skips have been expanded such that the LEFT and RIGHT buttons have one sensibility value in the reverse and forward directions respectively.  The DOWN and UP buttons have another sensibility value in the reverse and forward directions respectively.

Next, the actions of the NUMBER buttons is now configurable.  The "Numeric skip mode" setting has options of Symmetrical skips" where the number buttons work as they have always worked, "Defined skips" where each number can be assigned to skip any number of seconds in either direction, and "Percentage skips" where each number will jump to a percentage position within the media represented by that number.

The percentage skips work such that pressing 1 will jump to the 10% point, 6 to the 60% point, 0 to the 0% (start of media) point, etc.  If a second number is pressed quickly after the first button press then you can achieve a more granular jump. 5 followed by 8 will jump to the 58% point into the media file.

Another change to the new "SeekBar" screen is that if the media being played allows immediate skipping then the skip requested will be immediately actioned.  That will allow users to find the point of interest without having to exit the screen.

As per the old "Seekbar" screen, if the user presses EXIT the screen will close and the user will be returned to the point immediately prior to the "SeekBar" being activated.  Pressing OK will exit the screen and make the current position of the media be the same place selected in the "SeekBar" screen.

It is planned to extend these new media skip options to all core media playback screens within OpenATV.  It is envisioned that the "SeekBar" screen itself may be retired when these new skip options are available more widely.  We are working on bringing these new skip options to MovieSelection, Timeshift, CutListEditor and possibly other screens.

### DIFF
--- a/lib/python/Screens/InfoBarGenerics.py
+++ b/lib/python/Screens/InfoBarGenerics.py
@@ -32,6 +32,8 @@ from Components.TimerList import TimerList  # Deprecated!
 from Components.Timeshift import InfoBarTimeshift
 from Components.UsageConfig import defaultMoviePath, preferredInstantRecordPath, preferredTimerPath
 from Components.VolumeControl import VolumeControl
+from Components.Renderer.PositionGauge import PositionGauge
+from Components.Renderer.Progress import Progress
 from Components.Sources.Boolean import Boolean
 from Components.Sources.ServiceEvent import ServiceEvent
 from Components.Sources.StaticText import StaticText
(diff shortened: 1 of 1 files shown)


IanSav:
The issue reported by Sonar should be resolved as more of the code improvement and cleanup progresses.

PR #3638, opened 2025-10-22
[Picon.py] Exclude control character codes

AbuBaniaz:
Exclude control character codes in some channel settings.


jbleyel:
Are you sure ? Why not:

`name.replace('\xc2\x86', '').replace('\xc2\x87', '')`

PR #3666, opened 2025-12-23
[BackupRestore] Add user's ABM folders

AbuBaniaz:
Save/restore ABM user custom/provider files.


jbleyel:
If you add folders then add "/" at the end.

PR #3668, opened 2025-12-28
[SoftCSA]

jbleyel:
[SoftCSA] 
* introduce configurable settings
* introduce relevant information in SystemInfo


DimitarCC:
Where is the other part of softCSA code?
That PR is useless without the other code....

atvcaptain:
you can test first with openatv devel, after fix all issue you get the code

DimitarCC:
> you can test first with openatv devel, after fix all issue you get the code

I know that but this PR uses eDVBCSA which is not in this repo. So that PR is useless anyway now.

PR #3669, opened 2026-01-02
Enhance GPT slot management and boot config handling

ostende:
Added functionality to dynamically manage the number of slots based on the selected device's partition size. Updated boot configuration handling to accommodate the new slot management.


atvcaptain:
merge with some fixes

PR #3672, opened 2026-01-14
[VideoMode] Fix refresh Setup on changedEntry

GhostofGeeeee:
not only for keyLeft and keyRight


jbleyel:
This is better:
[VideoMode.py](https://github.com/user-attachments/files/24676224/VideoMode.py)

GhostofGeeeee:
Ok, but think about this wrong placed, useless line.

jbleyel:
[VideoMode.py](https://github.com/user-attachments/files/24692087/VideoMode.py)

PR #3675, opened 2026-01-19
[InstantRecordTimerEdit] remove wrong line

GhostofGeeeee:
[This commit](https://github.com/openatv/enigma2/commit/e5b09ccd90600db4d1c720c0752383a1199af545) starts removing the default margin(serious used for recordings).
Later it was replaced by the zap margin but now we have a seperate zap margin and this should not removed anymore.


jbleyel:
What is wrong with this line?
Please provide more details.

PR #3735, opened 2026-03-15
[Navigation.py] fix 8178f8eeaabb38a264cf0b81a794aeeef0dd61ec

Huevos:
In the original the import is "from ServiceReference import ServiceReference" but in OpenATV it is "import ServiceReference"


ghost:
Thank you, this solves the problem with this plugin.
https://github.com/DimitarCC/iptv-m3u-reader/issues/59#issuecomment-4062802564

PR #3739, opened 2026-03-20
Update supported WiFi encryption.

lukasz1992:
- Drop WEP and WPA, they do not match today standards
- Introduce option to use WPA3

### DIFF
--- a/lib/python/Plugins/SystemPlugins/WirelessLan/Wlan.py
+++ b/lib/python/Plugins/SystemPlugins/WirelessLan/Wlan.py
@@ -8,14 +8,12 @@ from Components.Network import iNetwork
 from Tools.Directories import fileReadLines, fileWriteLines
 
 MODULE_NAME = __name__.split(".")[-1]
-MODE_LIST = ["WPA/WPA2", "WPA2", "WPA", "WEP", "Unencrypted"]
-WEP_LIST = ["ASCII", "HEX"]
+MODE_LIST = ["WPA3", "WPA2", "Unencrypted"]
 
 config.plugins.wlan = ConfigSubsection()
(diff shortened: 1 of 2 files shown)


jbleyel:
Adding WPA3 is ok if all receivers supports this. Otherwise we need to add a filter for non supported devices. Removing the WEP and WPA is not ok because we don’t know if it’s still used even if it’s no longer state of the art.

jbleyel:
Update:

Many of the current used WiFi drivers don’t support WPA3. Especially the old ones.
New drivers needs a newer kernel which means many receivers won’t work.

How did you test your changes and what receiver did you use?

lukasz1992:
Zgemma with 4.4 kernel. I backported rtl8xxxu driver (that supports PMF) and loaded it. As wpa_supplicant supports SAE, I decided to give a try by manually editing wpa_supplicant.wlan0.conf.

Indeed it works correctly, wpa_cli reports SAE is being used.

Unfortunately it is hard to detect if WPA3 can work:
- simple test requires iw tool to be installed (only this command can realibly query driver support)
- wpa_cli can be used to check if installed version of wpa_supplicant supports SAE
- just testing the connection can help - you will see if wpa_supplicant stops immediately
- maybe the simplest way would be to define allowlist for wpa3

List of supported drivers that can do it:
- zd1211rw (already included in image - pre-ancient chip)
- carl9170 (already included in image - ancient chip)
- mt7601u (still used, needs a little patch https://github.com/torvalds/linux/commit/53c2cb8df3278481006255660ffc571ff4c5e0f1)
- rtl8xxxu and possibly more.
For all of mentioned, encryption is done in software - but many of CPUs are fast enough to handle >10MB/s traffic with tiny load increase

jbleyel:
We plan a complete refactoring of the network code and then we will add WPA3.

PR #3743, opened 2026-03-23
POC/WIP - EGL + GLES rendering

sreichholf:
I thought I'd use the bit of knowledge left and googles antigravity to create a prototype of an EGL/GLES 2/3 based rendering pipeline.
I tried to guide the AI to do the right things from the basic parts (and mistakes) I could remember from my time back in the day.

Here's the core concept I figured would make sense:
* All the "heavy lifting" should be done inside gEGLDC (EGL initialization, rendering with gles)
* Add a generic font atlas that should also work with 2D HW accel because there's no point in rendering with gles when we have to render every letter in software all the time. 
Currently there's no "font preloading" for standard skin fonts, we might want that in the future. For now it's populated on-demand, and therefore a bit more in danger of high fragmentation.
* It will not work without any widget adjustments at all, either due to required bug fixes or due to the wish of direct HW rendering support (see eCanvas) but they're fairly minimal for now.

I created the ticket so people can follow. 
This is a draft for a reason and currently has never even been tried to compile.

### DIFF
--- a/lib/gdi/Makefile.inc
+++ b/lib/gdi/Makefile.inc
@@ -49,7 +49,9 @@ gdiinclude_HEADERS = \
 
 if HAVE_EGL
 gdi_libenigma_gdi_a_SOURCES += \
-	gdi/egl/platform/amlogic/amlogic_window_provider.cpp
+	gdi/egl/egl_config.cpp \
+	gdi/egl/egl_init.cpp \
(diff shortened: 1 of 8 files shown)


jbleyel:
I have changed the branch.
We can merge to the extra branch if the code is ready for test.

PR #3744, opened 2026-03-25
[Information.py] Fix previous commit

IanSav:
- Restore the "time" import that is being used for "strftime" and "strptime".
- Sort the imports.


IanSav:
The correction is not correct.  ;)

PR #3760, opened 2026-05-04
subtitle: dump DVB subtitle frames to /tmp/current_sub.png

ahmedmoselhi:
using this so i can pump it to ocr service and translate it to my own language


jbleyel:
We already have subtitle translation.
So why this extra overhead?

jbleyel:
Sorry if this sounds rude, but this code is crap.
This code creates png files even if they are not used.
This will never committed this way.

PR #3763, opened 2026-05-06
[Harddisk.py] Modify automatic mounts

AbuBaniaz:
### Problem

On boxes with multiple storage devices, block devices were mounted at raw kernel-assigned paths (e.g. `/media/sdb1`, `/media/mmcblk1p1`) rather than the expected `/media/hdd`, `/media/usb`, `/media/mmc` etc. Users had to manually configure mountpoints via MountManager.

### Solution

This PR adds automatic mountpoint assignment and remounting during `enumerateBlockDevices`, following the same principles already used by `enumerateHotPlugDevices`.

### Changes

`addHotplugPartition` now calls `_getCorrectMountpoint` to determine the correct mountpoint. The logic is split into small focused helper methods:

- `_getDeviceUuid` — resolves a device's filesystem UUID via `/dev/disk/by-uuid`
- `_isDeviceInFstab` — checks fstab by device path or UUID
- `_getUsedMountpoints` — returns mountpoints already assigned in the current session
- `_remount` — unmounts stale mounts and remounts to the correct path
- `_getMmcMountpoint` — assigns `/media/mmc`, `/media/mmc1` etc. for external SD cards
- `_getUsbMountpoint` — assigns `/media/hdd`, `/media/usb`, `/media/usb2` etc. for USB/SATA
- `_getCorrectMountpoint` — dispatcher that calls the above based on device type
- `_unmountStaleExternalMmcMounts` — called synchronously at the start of `enumerateBlockDevices` to clean up stale `/media/` mounts of external MMC devices left over from a previous boot

### Behaviour

- First device (USB/SATA) → `/media/hdd`
- Additional USB/SATA devices → `/media/usb`, `/media/usb2` etc.
- External SD card with no other devices → `/media/hdd`
- External SD card with USB/SATA present → `/media/mmc`
- Any existing fstab entry (UUID or `/dev/`-based, whether user-configured or written by `enumerateHotPlugDevices`) is always respected and never overwritten
- No fstab entries are written by this code — that remains the sole responsibility of `enumerateHotPlugDevices`
- On dm900/dm920, `mmcblk0p3` is mounted as `/media/data` (ported from OpenViX, see openatv/enigma2#[ViX commit b26d953])

### Notes

- `_unmountStaleExternalMmcMounts` uses `subprocess.call` (synchronous) intentionally — it must complete before the enumeration loop begins
- The dm900/dm920 model check mirrors existing behaviour in OpenViX
- Tested on dm900 with internal HDD, two USB drives and an external SD card across multiple reboots and enigma2 restarts

### DIFF
--- a/lib/python/Components/Harddisk.py
+++ b/lib/python/Components/Harddisk.py
@@ -614,6 +614,7 @@ class HarddiskManager:
 	def enumerateBlockDevices(self):
 		print("[Harddisk] Enumerating block devices.")
 		black = BoxInfo.getItem("mtdblack")
+		self._unmountStaleExternalMmcMounts(black)
 		for blockdev in listdir("/sys/block"):
 			if blockdev.startswith(("ram", "rom", "loop", "zram", "md0", black)):
 				continue
@@ -677,7 +678,8 @@ class HarddiskManager:
 		error, blacklisted, removable, is_cdrom, partitions, medium_found = self.getBlockDevInfo(self.splitDeviceName(device)[0])
 		if not blacklisted and medium_found:
 			description = self.getUserfriendlyDeviceName(device, physdev)
-			p = Partition(mountpoint=self.getMountpoint(device), description=description, force_mounted=True, device=device)
+			mountpoint = self._getCorrectMountpoint(device)
+			p = Partition(mountpoint=mountpoint, description=description, force_mounted=True, device=device)
 			self.partitions.append(p)
 			if p.mountpoint:  # Plugins won't expect unmounted devices
 				self.triggerAddRemovePartion("add", p)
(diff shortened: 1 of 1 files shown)


jbleyel:
This code is spaghetti and very bad.

AbuBaniaz:
Can you please do it in a better way.

jbleyel:
> Can you please do it in a better way.

Yes, if you can explain the reason for this change.

AbuBaniaz:
Thanks.  The main reason is to expand on the automation of mounting the storage devices. This is nothing new, just extending the existing processes that apply for some receivers but not all. ATV mounts the internal flash memory as HDD as default but not the physical HDD.

1. Add mounting of the data partition on the dm900/dm920
2. Extend mounting of MMC cards as MMCx instead of SDbusLetterX, unless it's the only storage device. In which case mount as HDD. Honour any mountpoints assigned by the user which will use UUID and is more robust.
3. Extend mounting of USB devices as USBx instead of  SDbusLetterX, unless it's the only storage device. In which case mount as HDD. Honour any mountpoints set assigned by the user which will use UUID and is more robust.

jbleyel:
We do not automatically mount devices anymore.
This has been removed by purpose.

AbuBaniaz:
To remove, it would have had to exist in the first place. It never did.

jbleyel:
> To remove, it would have had to exist in the first place. It never did.

Oh yes .. the mount script has done that, and this has been completely changed since we switched from mdev to udev.

PR #3772, opened 2026-05-18
Picon, remove service type check and cosmetic

AbuBaniaz:
add a vix change and then a cosmetic change to better show selection process

### DIFF
--- a/lib/python/Components/Renderer/Picon.py
+++ b/lib/python/Components/Renderer/Picon.py
@@ -74,15 +74,33 @@ def getPiconName(serviceName):
 	fields = GetWithAlternative(serviceName).split(":", 10)[:10]  # Remove the path and name fields, and replace ":" by "_"
 	if not fields or len(fields) < 10:
 		return ""
+	full_ref = "_".join(fields)  # 1. Full service reference
+	partial_ref = "1_0_1_%s_0_0_0" % "_".join(fields[3:7])  # 2. SID, TSID, ONID and NAMESPACE with padding
+	partial_ignore_subnet = "1_0_1_%s_0_0_0" % ("_".join(fields[3:7])[:-4] + "0000")  # 3. SID, TSID, ONID and NAMESPACE with padding, ignoring namespace subnet differences
+	basenames = [full_ref, partial_ref, partial_ignore_subnet]
+
 	pngname = ""
-	basenames = ["_".join(fields), (p := "1_0_1_%s_0_0_0") % (x := ("_".join(fields[3:7]))), p % (x[:-4] + "0000")]
-	for basename in dict.fromkeys(basenames).keys():  # skip duplicates, maintain order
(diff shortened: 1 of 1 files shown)


lib/python/Components/Renderer/Picon.py
```diff
@@ -74,16 +74,11 @@ def getPiconName(serviceName):
...
-		fields[6] = fields[6][:-4] + "0000"  # Remove "sub-network" from namespace
-		pngname = findPicon("_".join(fields))
-	if not pngname and fields[0] != "1":
-		fields[0] = "1"  # Fallback to 1 for other reftypes
-		pngname = findPicon("_".join(fields))
-	if not pngname and fields[2] != "1":
-		fields[2] = "1"  # Fallback to 1 for services with different service types
-		pngname = findPicon("_".join(fields))
+	pngname = ""
+	basenames = ["_".join(fields), (p := "1_0_1_%s_0_0_0") % (x := ("_".join(fields[3:7]))), p % (x[:-4] + "0000")]
```
jbleyel: This will change the search order and probably break some picon packages.

Have you checked this?
Radio is not TV and HD is not SD but the sub network can be different.

That’s why ignoring subnetwork is the first fallback in the current code.

AbuBaniaz:
The existing code converts full namespace to partial namespace. So instead of  `1_0_1_1CF_B_1_13305C_0_0_0` it will look for `1_0_1_1CF_B_1_130000_0_0_0`. So it should be the last thing to change/ignore to ensure strictness for as long as possible. This has been the case for many months on Vix and PLI with no issues.  A pull request was not sent to ATV when that change was made/nor were you asked to pick it.

It won't break anything as no functional change has been made, just ordered in a more human readable form/list.

jbleyel:
> The existing code converts full namespace to partial namespace. So instead of `1_0_1_1CF_B_1_13305C_0_0_0` it will look for `1_0_1_1CF_B_1_130000_0_0_0`. So it should be the last thing to change/ignore to ensure strictness for as long as possible. This has been the case for many months on Vix and PLI with no issues. A pull request was not sent to ATV when that change was made/nor were you asked to pick it.
> 
> It won't break anything as no functional change has been made, just ordered in a more human readable form/list.

The old order is: 
1. full sref
2. sref with normalized namespace
3. sref with 1:0:1

The new order is:
1. full sref
2. sref with 1:0:1
3. sref with normalized namespace

This will break picons like 1:0:2 or 1:0:19 with different subnetwork.

AbuBaniaz:
OK, sorry about that. I had used the PLI/Vix order.

jbleyel:
> OK, sorry about that. I had used the PLI/Vix order.

https://github.com/OpenViX/enigma2/commit/e0b6371f365f65039f70953f7b37a531497c028f

The original order on Vix was the same as ATV.

jbleyel:
Your last change is still wrong.

AbuBaniaz:
Ok, no worries. I'll close this.

PR #3781, opened 2026-05-31
[InfoBarGenerics] tolerate non-ASCII paths on setResumePoint

WXbet:
posixpath.realpath() uses fsencode under ASCII locale → UnicodeEncodeError on filenames with chars like en-dash (U+2013), killing the MoviePlayer when the user presses STOP. Swallow it and skip the dead-cache cleanup for that entry. Seen on aarch64 devices (e.g. dreamone)

### DIFF
--- a/lib/python/Screens/InfoBarGenerics.py
+++ b/lib/python/Screens/InfoBarGenerics.py
@@ -105,10 +105,13 @@ def setResumePoint(session):
 				for k, v in list(resumePointCache.items()):
 					if v[0] < lru:
 						candidate = k
-						filepath = realpath(candidate.split(":")[-1])
-						mountpoint = findMountPoint(filepath)
-						if ismount(mountpoint) and not exists(filepath):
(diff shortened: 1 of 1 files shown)


IanSav:
Can you please explain this issue in a bit more detail.  If there are places where Unicode characters are causing issues we should look to fix the problem rather than ignore it.  I don't understand where or why an en-dash is causing a problem.

WXbet:
Seems it only crashes on pressing STOP in mediaplayers on dreamone|two (aarch64) without this correction.

IanSav:
I don't have access to that hardware to debug the issue.  I would like to avoid specific hardware custom fixes whenever possible.  If you can ever find the root cause of this issue it would be fantastic if you could update the fix.

WXbet:
[20260531-145000-enigma2-crash.log](https://github.com/user-attachments/files/28476407/20260531-145000-enigma2-crash.log)
This is the corresponding crashlog.

PR #3807, opened 2026-06-23
MultiBootManager: refactor action maps, methods, and slot status handling

jbleyel:
MultiBootManager: refactor action maps, methods, and slot status handling

MultiBoot.py:
- Add detection of "hidden" slot status: if "usr/bin/enigma2x.bin" is
  present in the image directory (a renamed/disabled enigma2 binary),
  the slot is classified as "hidden" with display name "Disabled" instead
  of falling through to "Empty". This correctly distinguishes between a
  slot that was intentionally disabled and one that is genuinely empty.

MultiBootManager.py:
- Rename action maps to better reflect their purpose:
  · "deleteActions"   → "emptyActions"   (covers wipe, delete, and disable)
  · "moreSlotActions" → "addActions"
  · "editActions"     → "renameActions"
- Add "key" prefix to all key-handler methods to align with the
  established enigma2 naming convention:
  · cancel() / closeRecursive() → keyCancel() / keyCloseRecursive()
  · reboot()                    → keyReboot()
  · deleteImage()               → keyEmptySlot()
  · restoreImage()              → keyRestoreSlot()
  · moreSlots()                 → keyAddSlots()
  · keyEdit()                   → keyRenameSlot()
- Convert all standalone callback methods into closures nested inside
  their respective key-handler: rebootCallback, deleteImageCallback,
  deleteImageAnswer, disableChkrootAnswer, disableChkrootCallback,
  restoreImageCallback, renameSlotCallback, and renameCallback are no
  longer class-level methods.
- Rename getImagesList() to getSlotList() and nest its callback as an
  inner function; rename internal variables for clarity:
  · imageList / imageLists → slots / slotData
  · currentMsg / current   → activeMsg / active
- Replace all self["slotlist"].l.getCurrentSelection()[0] calls with
  the cleaner self["slotlist"].getCurrent()[0] API.
- Enable previously commented-out pageUp / pageDown navigation actions
  and add the corresponding keyPageUp() / keyPageDown() methods.
- Add dedicated "hidden" branch in selectionChanged(): hidden slots
  show only the BLUE/Restore action. Previously "empty" handled this
  role; "empty" now correctly maps to no available actions.
- Extract repeated _("MultiBoot Manager Actions") string into a local
  variable actionDescription to avoid redundant translation calls.
- Remove simple=True from MessageBox call in keyEmptySlot() to use the
  standard button layout.
- FIX: set new limits by using the new function updateLimits

config.py:
- ConfigSequence add new function for update limits

Information.py:
- use 'Active' instead of 'Current' slot

### DIFF
--- a/lib/python/Components/config.py
+++ b/lib/python/Components/config.py
@@ -1166,6 +1166,13 @@ class ConfigSequence(ConfigElement):
 			self.endNotifier = []
 		self.endNotifier.append(notifier)
 
+	def updateLimits(self, limits):
+		if not isinstance(limits, list) and len(limits[0]) != 2:
+			raise TypeError("[Config] Error: Limits must be [(min, max), ...] tuple-list!")
+		self.limits = limits
+		self.blockLen = [len(str(x[1])) for x in limits]
+		self.totalLen = sum(self.blockLen) - 1
+
 
 class ConfigCECAddress(ConfigSequence):
 	def __init__(self, default, auto_jump=False):
--- a/lib/python/Screens/Information.py
+++ b/lib/python/Screens/Information.py
@@ -941,7 +941,7 @@ class MultiBootInformation(InformationBase):
 		if self.slotImages:
 			slotCode, bootCode = MultiBoot.getCurrentSlotAndBootCodes()
 			slotImageList = sorted(self.slotImages.keys(), key=lambda x: (not x.isnumeric(), int(x) if x.isnumeric() else x))
-			currentMsg = f"  -  {_('Current')}"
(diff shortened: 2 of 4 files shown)


IanSav (changes requested):
The matching changes in Information.py should also be in this pull request.  (Information now also uses "Active" to describe the active slot rather than "Current".)

jbleyel:
> The matching changes in Information.py should also be in this pull request. (Information now also uses "Active" to describe the active slot rather than "Current".)

DONE

PR #3813, opened 2026-06-24
MultiBootManager: fix slot renaming for legacy images

WXbet:
Problem: rename had no visible effect on slots detected without an existing enigma.info file (e.g. legacy images). The slot list kept showing the derived name.

Cause: readSlotInfo(path) reads <path>.info plus <path>.conf, but only the isfile(infoFile) branch in analyzeSlot calls it. Slots that fall into the versionFile/enigma2-binary elif branches never trigger the .conf override.

Fix: touch an empty enigma.info as marker on rename so analyzeSlot takes the isfile(infoFile) branch and readSlotInfo merges enigma.conf as override. compiledate is pinned in enigma.conf alongside the marker, reusing the value analyzeSlot already inferred for the slot, so _formatSlotName keeps the "(YYYY-MM-DD)" suffix. Reset drops both the marker and the pinned compiledate.

Also: avoid the second mount/unmount round-trip on dreamone — the bootmanager section name for /data/bootconfig.txt is now read from the still-mounted image dir at the end of _renameSlotMounted and handed to updateDreamBootSection via a new sectionName= argument. The shared logic lives in a new _readSectionNameFromImageDir() helper, which getDreamBootSectionName() now uses too.

### DIFF
--- a/lib/python/Tools/MultiBoot.py
+++ b/lib/python/Tools/MultiBoot.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 from glob import glob
 from hashlib import md5
-from os import listdir, mkdir, rename, rmdir, stat
+from os import listdir, mkdir, rename, rmdir, stat, unlink
 from os.path import basename, exists, isdir, isfile, ismount, join, islink, realpath
 from re import search
 from struct import calcsize, pack, unpack, error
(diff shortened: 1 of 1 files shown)


IanSav:
Have all these recent changes been fully tested?  Testing and fixing should be done offline so that the commit history is clear and understandable.

This change, for example, concerns me because old image don't have an "enigma.info" and we shouldn't be adding one to that old image if it is not used by that image.  Creating an empty, and hence invalid, "enigma.info" is a very bad idea, as it may create other, unexpected, issues elsewhere.  I also think that the MultiBoot code should not be writing anything to any of the other slots or images that are not the active image.  The other slots should be accessed for reading only.  Perhaps, when I wrote this code, I should have made the mounts read-only to highlight that the other slots should never be changed.

Can you please consider developing and testing your code locally before committing unstable code to the main repository branches.  This whole stream of MultiBoot changes is now a mix of different and unrelated changes that are all mixed up so it is no longer possible to revert a single functional change.  For example, if the new rename actions was separated from the other changes we could have reverted it, in isolation, while the code was fixed.  Similarly, if all the Dreambox changes were in their own thread they could have been separated from the rename changes.

WXbet:
Hi Ian, of course I test every commit locally before even considering creating a PR. However, I can't cover every use case on my own and rely on feedback from the community.

To be honest, I’m not a fan of write access to other slots either; a while back, I even suggested handling slot renaming via the STARTUP files. In my opinion, that would be the cleanest approach. After all, users should be able to name the slots however they like.

IanSav:
Hi WXbet, when I first wrote all this MultiBoot related code I gave it to @atvcaptain to test on his vast collection of different boxes to ensure that all the combinations worked.  I have very limited hardware and was unable to test most conditions myself.  Mind you, I often patched my own boxes to simulate many of the conditions seen on hardware that I don't own.  Again, @atvcaptain was indispensable in providing knowledge about the other hardware out there.  This resulted in a clean commit with minimal correction required after the code was released.

You will note that I did a code review your changes to MultiBootManager to optimise the code and brought it up to current standards.  I believe it is still easy to read and maintain.

PR #3843, opened 2026-08-09
[OScamInfo] fix empty main screen on non-ISO timestamp formats

WXbet:
Some OSCam builds emit starttime/login as DD.MM.YYYYTHH:MM:SS+ZZZZ instead of ISO-8601. datetime.fromisoformat() raised ValueError, which was swallowed by callInThread, leaving the main screen blank while yellow/blue still worked. Add parseTimestamp() with a fallback to the DD.MM.YYYY format.


WXbet:
Thank you @jbleyel . Cann you please merge it into 8.0 branch too?

PR #3856, opened 2026-08-19
[skin] add more skin formula operators

xcentaurix:
currently the only operators that can be used in formulas are + and -, like e-20, e+40.
other operators can't be used because all values are scaled equally but values after e.g. * or / must not be scaled.
this change distiguishes between values that need to be scaled and those that have to be left unscaled.

this code was generated with claude code ai.

i unit tested it with the following skin:

		<widget font="Regular;20" halign="center" name="key_red" position="0,e-48" size="e/4,38" foregroundColor="keyRed" transparent="1" zPosition="2"/>
		<widget font="Regular;20" halign="center" name="key_green" position="1*e/4,e-48" size="e/4,38" foregroundColor="keyGreen" transparent="1" zPosition="2"/>
		<widget font="Regular;20" halign="center" name="key_yellow" position="2*e/4,e-48" size="e/4,38" foregroundColor="keyYellow" transparent="1" zPosition="2"/>
		<widget font="Regular;20" halign="center" name="key_blue" position="3*e/4,e-48" size="e/4,38" foregroundColor="keyBlue" transparent="1" zPosition="2"/>

### DIFF
--- a/lib/python/skin.py
+++ b/lib/python/skin.py
@@ -411,6 +411,9 @@ def parseColor(value, default=0x00FFFFFF):
 #
 def parseCoordinate(value, parent, size=0, font=None, scale=(1, 1)):
 	def scaleNumbers(coordinate, scale):
+		# Digits adjacent to "*" or "/", or immediately followed by "%", "w" or "h", are ratio/coefficient
+		# constants (e.g. the "4" in "e/4", the "3" in "3*e", the "25" in "25%") and must be left unscaled -
+		# only additive pixel offsets (e.g. the "48" in "e-48") represent real pixel quantities that need scaling.
 		inNumber = False
 		chars = []
 		digits = []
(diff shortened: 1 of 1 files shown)


IanSav:
Your suggestion and contribution is interesting.

Can you give us some more specific examples of what you want to achieve.  Please show the mathematical calculations and assumptions that you are making.  I would like to know what situations you see that the current system does not address.

As a further suggestion, rather than assuming what numbers should or should not be scaled based on their position in the line or their relationship to a mathematical operator I would suggest that all numbers be scaled except numbers that are specifically marked to not be scaled.  If you know Excel or any other spreadsheet you should know what I am suggesting.  Adding a character like "=" in front of a number should tell the code to NOT scale that number.

xcentaurix:
the main application is to divide a distance into equal distances for positioning like buttons or other items on different screen sizes. this avoids manual calculation of sizes and positions. the example i showed in the pr is such a key length/position calculation.
the equal sign could be done, but i cant think of an application where you would want to scale e.g. dividend and divisor.

PR #3858, opened 2026-08-20
[draft] Add ECMs and EMMs to the pids filters

catalinii:
This allows CIs to be used in a satip adapter and decrypt locally on VU+


jbleyel:
Please pull only to master branch.

Issue #3862, opened 2026-08-24
[Bug]: OpenATV 7.6 – BUG with DVB subtitles on some TV channels


yos825:
Hello,
I have found a reproducible DVB subtitle problem with OpenATV 7.6 on an Octagon SF8008 V3 Supreme decoder, receiving Hotbird 13°E.
On channels such as RAI 1, RAI 2 and RAI 3, SRF Info HD, TVP Polonia, subtitles selected with the SUBT button do not appear. These channels also broadcast Teletext subtitles (e.g. page 777).

DVB subtitles work correctly on channels I tested which do not broadcast TXT 777 subtitles, for example ARTE, TV5 Monde, TVE Internacional and TVP Info HD.

The most important test: the seller helped me install OpenATV 6.4 in another multiboot slot of the same Octagon receiver.

OpenATV 6.4: DVB subtitles work correctly on all tested TV channels.
OpenATV 7.6: DVB subtitles can be selected with the SUBT button, but are not displayed on the affected channels.

Therefore the hardware, satellite signal and receiver are exactly the same; only the OpenATV version changes.
I also tested “Prefer graphical DVB subtitles” in OpenATV 7.6 with both NO and YES — it makes no difference. In OpenATV 6.4 this option is set to NO, and the subtitles still work correctly.

I will be happy to provide additional technical information and perform further tests if required.

I hope this subtitle bug can be fixed in the latest OpenATV 8.x final version, if it has not already been corrected there.

Best regards,
Yos

PR #3863, opened 2026-08-25
[ServiceDVB] fix teletext subtitles never being displayed

WXbet:
Two independent defects made EBU teletext subtitles unusable on channels that only carry them (RAI 1/2/3, SRF info, TVP Polonia on 13E), while channels with real DVB subtitles were unaffected. Reported as #3862.

1) newSubtitlePage() only queued pages whose PTS was ahead of the decoder
   PTS and silently dropped everything else. Broadcasters that send the
   teletext PES without a PTS end up with m_pts = 0, so every page was
   discarded and nothing was ever displayed. Clamp pages without usable
   timing to the current PTS instead, which is what the code did before
   the timing rework, and keep the ordering and lifespan checks.

2) The first eventNewProgramInfo after a zap can come from the service
   cache, which often has no cached text pid. The teletext parser was
   then started with pid -1 and never set up a PES filter, and since the
   start call sits behind the mustPlay gate the real pid from the PMT
   100ms later was never picked up. Restart the parser whenever a later
   program info brings a text pid that differs from the running one.

### DIFF
--- a/lib/dvb/teletext.h
+++ b/lib/dvb/teletext.h
@@ -50,6 +50,7 @@ public:
 	static const int max_id = 26;
 	static const char *const my_country_codes[];
 	int start(int pid);
+	int getPid() const { return m_pid; }
 	void setPageAndMagazine(int page, int magazine, const char *lang);
 	void setMagazine(int magazine);
 	void connectNewStream(const sigc::slot<void()> &slot, ePtr<eConnection> &connection);
--- a/lib/service/servicedvb.cpp
+++ b/lib/service/servicedvb.cpp
@@ -1385,6 +1385,16 @@ void eDVBServicePlay::serviceEvent(int event)
 					}
 				}
 			}
+			else if (m_teletext_parser && m_decoder_index == 0)
+			{
+				// same as in updateDecoder: the text pid may only show up in a later program info
+				eDVBServicePMTHandler::program program;
+				if (!m_service_handler.getProgramInfo(program) && program.textPid >= 0 && program.textPid != m_teletext_parser->getPid())
+				{
+					m_teletext_parser->start(program.textPid);
+					eDebug("[eDVBServicePlay] Teletext parser restarted on PID %04x", program.textPid);
+				}
+			}
 
 			// Late-start case: session is active but SoftDecoder may not be running yet
 			// This happens when we switched to a channel where CSA-ALT was already cached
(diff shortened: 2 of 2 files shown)


yos825:
> Two independent defects made EBU teletext subtitles unusable on channels that only carry them (RAI 1/2/3, SRF info, TVP Polonia on 13E), while channels with real DVB subtitles were unaffected. Reported as #3862.
> 
> 1. newSubtitlePage() only queued pages whose PTS was ahead of the decoder
>    PTS and silently dropped everything else. Broadcasters that send the
>    teletext PES without a PTS end up with m_pts = 0, so every page was
>    discarded and nothing was ever displayed. Clamp pages without usable
>    timing to the current PTS instead, which is what the code did before
>    the timing rework, and keep the ordering and lifespan checks.
> 2. The first eventNewProgramInfo after a zap can come from the service
>    cache, which often has no cached text pid. The teletext parser was
>    then started with pid -1 and never set up a PES filter, and since the
>    start call sits behind the mustPlay gate the real pid from the PMT
>    100ms later was never picked up. Restart the parser whenever a later
>    program info brings a text pid that differs from the running one.

Hello WXbet and jbleyel,

Thank you very much for investigating this issue and implementing the fix.

Before I test it on my Octagon SF8008, could you please confirm whether you were able to reproduce the problem on RAI 1/2/3 on Hotbird 13°E with OpenATV 7.6, and whether you tested these channels again after applying the fix?

In particular, I would like to know if SUBT → 777 Italian now displays the subtitles correctly on the latest OpenATV build.

If this has been confirmed, could you please tell me which OpenATV version/build already contains the correction and how I can install or update to it for testing?

Thank you again for your help.

Best regards,
Yos

yos825:
Hello WXbet and jbleyel,

Thank you very much for investigating this issue and implementing the fix.

Before I test it on my Octagon SF8008, could you please confirm whether you were able to reproduce the problem on RAI 1/2/3 on Hotbird 13°E with OpenATV 7.6, and whether you tested these channels again after applying the fix?

In particular, I would like to know if SUBT → 777 Italian now displays the subtitles correctly on the latest OpenATV build.

If this has been confirmed, could you please tell me which OpenATV version/build already contains the correction and how I can install or update to it for testing?

Thank you again for your help.

Best regards,
Yos

jbleyel:
You need to install ATV 8.0 and then the devel feed.

PR #3868, opened 2026-08-27
[DAB+] * fix decoder, TS adapter and audio path defects

WXbet:
Started as a Quality Gate fix and grew into a review of the DAB+ code, then into a no audio report on a pulse4kmini. Seven changes.

### Audio output

**`[eServiceDAB]` feed LOAS to sinks that advertise it**

On a pulse4kmini DAB+ produced no sound at all, with or without the other commits in this PR, so this is not a regression from them. The scan, FIC, MSC, PAD, DLS and SLS all worked, the audio queue simply overran once a second and leaked. Measured standalone with enigma2 stopped, so nothing else held the decoder:

| pipeline | material | result |
|---|---|---|
| LOAS → dvbaudiosink | 120 KB ≈ 9.2 s | plays, ends after 9.8 s, real time |
| PCM → dvbaudiosink | 819 KB ≈ 4.3 s | no buffer consumed within 30 s |
| PCM → fakesink | same buffers | done in 4 ms |

`dvbaudiosink` advertises `audio/x-raw` and does not consume it, and the DAB pipeline ended in `faad ! audioconvert ! audioresample ! dvbaudiosink`. `emitLOAS` already built a LOAS stream and only used it for the diagnostic capture; `aacparse` reads that capture back as `audio/mpeg mpegversion=4 rate=48000 channels=2 stream-format=loas`, so it is well formed. The 9.8 s also match the 960 sample transform, at 1024 the same data would have taken 10.5 s, so the hardware decoder handles DAB+ correctly here.

The path is chosen from the sink template naming `loas` in `stream-format`, not from a caps query: `dreamaudiosink` leaves `stream-format` open, which matches any value, and then maps `mpegversion=4` to `AV_CODEC_ID_AAC` without looking at the format, so a query would have switched it too and it would decode LATM as plain AAC. Since a template is a claim and not a guarantee, three queue overruns within eight seconds re-tap with the software decoder and the rejection is remembered for the session.

Note for reviewers with other hardware: this covers a sink that consumes nothing. A decoder that takes the buffers and renders them wrong, for instance by treating the 960 sample transform as 1024, would sound 6 to 7 % too low and too slow and can only be caught by listening.

### Correctness

**`[dabtsadapter]` endless loop when ETI multiframe alignment fails**
`processTSNA` advances the bit position and clears the sync flag when the block phase or the multiframe cannot be validated, but `findE1Sync` always searched from bit 0 and overwrote the position with the same result, so the same candidate failed again forever. One corrupted management byte is enough, `validMultiframe` requires all 24 block and superframe numbers to be consecutive. The worker thread then spins at 100% and never checks `m_stop` again, which also blocks the join when the service is stopped. Affects `dab://ts2na12` and `dab://ts2na`.

**`[dabdecoder]` AF packet length overflowed `size_t`**
`processAF` read the 32-bit LEN field and added the header and CRC overhead before comparing against the available length. On a 32-bit target that addition wraps, so a LEN close to `0xffffffff` produced a small `packetLength`, passed the bound and handed the unmodified LEN to `processTags`, which then walked far past the buffer. The data comes from the transport stream, so it is not trusted input.

**`[dabdecoder]` expected MST end derived from FL was one word too high**
EN 300 799 counts FL in 32-bit words from the FC field, excluding SYNC, EOF and TIST, so the MST ends at `4 + FL * 4`. `feedETI` expected `8 + FL * 4`. With the offsets `feedETI` uses itself, `position` ends at `12 + 4 * NST + MST`, which equals `4 + FL * 4` for a conformant frame, so the old constant could never match and every direct ETI frame was counted as a CRC error and dropped. This affects the three TS transports, which reach the decoder through `feedETI`; the MPE/EDI path goes through `feedEDI` and is unaffected.

**`[dabdecoder]` MSC stream selected by start address**
`feedMSC` skipped a stream only when the sub-channel id and the start address both differed, so a stream of a different sub-channel was accepted whenever its start address happened to match the selected one. SubChId identifies the sub-channel and the start address comes from the same FIG 0/1 record, so matching on the id alone is both sufficient and safe across a reconfiguration.

### Static analysis

**`[dabdecoder]` LATM bit ranges made provable**
SonarCloud fails the Quality Gate on New Code with `cpp:S3949` in `LATMBitWriter::addBits`: it cannot prove that `m_byte_bits` stays in [0,7], so it assumes `8 - m_byte_bits` underflows, `copyBits` becomes `count` and `0xff >> (8 - copyBits)` shifts by `(size_t)-3`. At runtime this cannot happen, the modulo keeps the member in range, so it is a false positive. Masking the member makes the range local and provable without changing the generated bitstream. Also clears `cpp:S1905` and `cpp:S3230` in the same class.

### Translations

**`[DABScan]` dab.xml parser diagnostics out of the GUI**
The parser raised a separate translated message for every way a feed definition can be wrong and joined them into the transponder line. Those texts only help someone writing a `dab.xml`, the shipped file is valid. They are now plain untranslated text in the debug log. Eight translatable strings less, none added. This also corrects the message shown when the file is readable but its feed entries are invalid, which claimed the satellite positions were unsuitable.

### Verification

- LATM: old and new `addBits` compared over the full parameter space of `buildAudioSyncStream` plus 200000 randomised bit sequences, byte identical output, clean under `-fsanitize=undefined`
- TS adapter: a harness feeding a buffer that carries the E1 sync pattern but no valid block sequence does not return before the fix and returns after it
- ETI: a frame built to EN 300 799 with both CRCs is rejected as a CRC error before the fix and accepted after it
- Audio: on air on Astra 19.2E, DAB Bundesmux 1 and 2, several services across three bit rates and sub-channels. Queue overruns went from 402 to 0, the software decode disappeared from the pipeline, the fallback never triggered, no GStreamer errors
- Built for an armv7 target (pulse4kmini, openatv 8.0) with no new compiler warnings

### DIFF
--- a/lib/python/Screens/DABScan.py
+++ b/lib/python/Screens/DABScan.py
@@ -49,7 +49,7 @@ class DABScan(ServiceScan):
 
 	def runScan(self):
 		if not self.loadFeeds():
-			message = _("No supported DAB+ feeds are available for the satellite positions configured on this box.") if self.knownFeeds else _("No valid DAB+ feeds configured. Check /etc/enigma2/dab.xml.")
+			message = _("No supported DAB+ feeds are available for the satellite positions configured on this box.") if self.knownFeeds and not self.feedErrors else _("No valid DAB+ feeds configured. Check /etc/enigma2/dab.xml.")
 			self.finishWithError(message)
 			return
 		if self.session.nav.RecordTimer.isRecording():
@@ -71,7 +71,7 @@ class DABScan(ServiceScan):
 		try:
 			root = parse(xmlPath).getroot()
 			if root.tag != "dabFeeds":
-				raise ValueError(_("Invalid DAB feed file root element"))
+				raise ValueError("invalid root element")
 			definitions = [(node, None, None) for node in root.findall("feed")]
 			for satellite in root.findall("satellite"):
 				for transponder in satellite.findall("transponder"):
@@ -91,14 +91,14 @@ class DABScan(ServiceScan):
 						continue
 					self.feeds.append(feed)
 				except (TypeError, ValueError) as err:
-					feedName = node.get("name") or node.get("id") or _("Unknown feed")
+					feedName = node.get("name") or node.get("id") or "unknown feed"
 					self.feedErrors.append(f"{feedName}: {err}")
 		except (OSError, ParseError, ValueError) as err:
 			self.feedErrors.append(f"{xmlPath}: {err}")
 		if self.knownFeeds:
 			print("[DABScan] Feed definitions: %d known, %d selected, %d unavailable, %d unsupported, %d disabled." % (self.knownFeeds, len(self.feeds), self.skippedUnavailableFeeds, self.skippedUnsupportedFeeds, self.skippedDisabledFeeds))
-		if not self.feeds and not self.feedErrors:
-			self.feedErrors.append(_("%d feeds are on unavailable satellites, %d use unsupported transports and %d are disabled") % (self.skippedUnavailableFeeds, self.skippedUnsupportedFeeds, self.skippedDisabledFeeds))
+		for error in self.feedErrors:
+			print(f"[DABScan] {error}")
 		return bool(self.feeds)
 
 	def feedAttribute(self, node, transponder, satellite, attribute, default=None):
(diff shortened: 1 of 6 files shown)


atvcaptain (approved):
all final test ok

StreamForge83:
After this commit, it stopped working completely on the Vu+ Zero 4K. It doesn't scan for DAB bouquets, and the DAB bouquets generated yesterday have also stopped working.
Astra 28.2
fail
f1
f2

atvcaptain:
Thank you for the report.
The playback problem is caused by the old dvbaudiosink. Please update to the latest complete image, including the updated GStreamer DVB media sink, before testing playback again.
The scan failure is a separate issue. Please:
1. Enable verbose/all Enigma2 debug logging.
2. Restart Enigma2.
3. Run the DAB+ scan for Astra 28.2°E.
4. Let both BBC National and D1 National finish or time out.
5. Attach the complete, unedited Enigma2 debug log—from before opening the DAB+ scan until the final result appears.
Please also provide the image build date and tell us whether /etc/enigma2/dab.xml exists, because that file overrides the packaged DAB feed definitions.
Screenshots or filtered log excerpts are not sufficient because the relevant tuner, PID tap, TS/NA and FIC messages may appear earlier in the log.

WXbet:
@StreamForge83 : Check latest Image from today. It contains this fix: https://github.com/openatv/enigma2/pull/3870/changes/7678477429342416c9728f8307de515182d88dae

StreamForge83:
DAB works correctly on the 8.0 beta. Something must have gone wrong with the DEVEL build on my receiver, so I’ll wait until the DEVEL images are rebuilt and then check again. Thank you.

WXbet:
Try 8.0 beta from today, all the things from devel were merged today. The fix above is based on your feedback here and part of the changes. If it still not work we need a debug log from scanning.

StreamForge83:
After today’s DEVEL update, DAB+ is working correctly now. Thank you.

PR #3869, opened 2026-08-27
[EPGCache] search 0x4D description text, truncate results fairly

### LINKED ISSUE #2491: [EPGCache] Can't search description with Freeview/Virgin providers selected
While testing OpenWebif's EPG search, I've found that searching EPG by description (using OpenWebif) returns no results.

Freesat w/ Virgin actiive returns no results, however...

With OpenTV EPG activated and channel tuned for a while, the same text-in-description search then returns the expected results.

I believe EPGCache populates shortDescription or longDescription based on activated EPG providers, and that the search is only performed on one of those fields.

WXbet:
Fixes #2491

### 1. The description search never looked at the short event descriptor text

`PARTIAL_DESCRIPTION_SEARCH` only matched against the extended event descriptor (0x4E). The short event descriptor (0x4D) carries **two** strings — `event_name` and `text` — and the second one holds the description for providers that broadcast no 0x4E at all.

The cache already stores that text: `eventData::eventData()` splits an incoming 0x4D into a title descriptor (`text_length == 0`) and a text descriptor (`event_name_length == 0`), so both end up in `descriptors` as separate 0x4D entries. The search simply never read the second one.

The reported consequence: with Freesat/Virgin active, a description search through OpenWebif returns nothing, while the same query works once OpenTV EPG has filled in extended descriptors. Displaying the description works either way — only the search is blind to it.

The search now accepts both tags and reads `text_length` at `data[6 + event_name_length]` with the text behind it, using the same codepage handling as the title search. Both offsets are bounds-checked against `descriptor_length`, because `event_name_length` is unvalidated when descriptors are restored from `epg.dat`.

### 2. `maxmatches` truncated along hash order

The result phase stopped as soon as the budget was spent, mid-traversal over `eventDB` — an `unordered_map`, so service order is plain hash order. Whichever services hashed first consumed the entire budget, and matches from every other source were dropped even though they had already been found. A source with many events can crowd out all the others, which to the user looks like "the search finds nothing".

Matches are now collected first — start time, service and event pointer only, no Python objects — then sorted by start time and cut to `maxmatches`. Reference expansion and tuple building stay bounded by the budget, so only the collection pass does extra work.

**Behaviour change:** results come back in chronological order instead of hash order. The in-tree callers (`EpgList.py`, `EventView.py`) do not rely on the previous grouping.

Also in that commit:

- Reference expansion honours the budget now. It used to `--maxcount` past zero, and with `maxcount` negative the enclosing `while (… && maxcount)` stayed true, so the traversal ran on to the end of the database.
- Descriptor CRCs are looked up in a hash set instead of scanning the deque per event, since the traversal no longer aborts early.
- Dropped the base-service-first machinery in `SIMILAR_BROADCASTINGS_SEARCH`. It has been dead since 30ee81509d ("Refactor eEPGCache", 2021): that commit turned the assignment `refstr = PyString_AS_STRING(obj)` into a declaration `const char *refstr = …` inside the branch, shadowing the outer variable. The outer `refstr` has been `NULL` ever since, so `ref` was never valid and `if (first)` unreachable. Its purpose was to secure the base service a place in a scarce budget — which collecting every hit up front makes unnecessary.

### DIFF
--- a/lib/dvb/epgcache.cpp
+++ b/lib/dvb/epgcache.cpp
@@ -2770,14 +2770,35 @@ PyObject *eEPGCache::search(ePyObject arg)
 						for (DescriptorMap::iterator it(eventData::descriptors.begin());
 							it != eventData::descriptors.end(); ++it)
 						{
-							uint8_t *data = it->second.data;
-							
-							eit_extended_descriptor_struct *extended_event_descriptor = (eit_extended_descriptor_struct *) ((u_char *) data);
-							if ( (u_char)extended_event_descriptor->descriptor_tag == (u_char)EXTENDED_EVENT_DESCRIPTOR ) // extended event descriptor
+							const uint8_t *data = it->second.data;
+							uint8_t descriptor_tag = data[0];
+
+							// the description can live in an extended event descriptor, or in the text
+							// field of a short event descriptor when no extended one is broadcast
+							if ( descriptor_tag == (u_char)EXTENDED_EVENT_DESCRIPTOR || descriptor_tag == (u_char)SHORT_EVENT_DESCRIPTOR )
 							{
-								int content_len = data[EIT_EXTENDED_EVENT_DESCRIPTOR_SIZE+1]; //struct extended_event_descriptor+item information (always "0", see epg.dat for structure)
-								const char *contentptr = (const char*)&data[EIT_EXTENDED_EVENT_DESCRIPTOR_SIZE+2];
-								if (data[EIT_EXTENDED_EVENT_DESCRIPTOR_SIZE+2] < 0x20) //Codepage
+								int content_len;
+								const char *contentptr;
+								if ( descriptor_tag == (u_char)EXTENDED_EVENT_DESCRIPTOR )
+								{
+									content_len = data[EIT_EXTENDED_EVENT_DESCRIPTOR_SIZE+1]; //struct extended_event_descriptor+item information (always "0", see epg.dat for structure)
+									contentptr = (const char*)&data[EIT_EXTENDED_EVENT_DESCRIPTOR_SIZE+2];
+								}
+								else
+								{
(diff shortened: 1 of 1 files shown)


AbuBaniaz:
My tests with the modified binary created for me by @WXbet are as attached. Many thanks. I have yet to set timers etc.


# EPG Search Test Results

Search type key:
- **title** (`partial match` / `PARTIAL_TITLE_SEARCH`) - searches event
  titles only.
- **description** (`partial description` / `PARTIAL_DESCRIPTION_SEARCH`)
  - searches description text.

## Query: "benson"

### Freeview + OpenTV

| Image/Binary | Search type | Result count | 28.2E count | DVB-T count |
|---|---|---|---|---|
| ATV 8.0 stock | 2 (partial) | 0 | 0 | 0 |
| wxbet modified | 2 (partial) | 0 | 0 | 0 |
| ATV 8.0 stock | 5 (partialdes) | 75 | 75 | 0 |
| wxbet modified | 5 (partialdes) | 89 | 74 | 15 |

### Freeview only

| Image/Binary | Search type | Result count | 28.2E count | DVB-T count |
|---|---|---|---|---|
| ATV 8.0 stock | 2 (partial) | 0 | 0 | 0 |
| wxbet modified | 2 (partial) | 0 | 0 | 0 |
| ATV 8.0 stock | 5 (partialdes) | 0 | 0 | 0 |
| wxbet modified | 5 (partialdes) | 16 | 0 | 16 |

## Query: "Law & Order: Special Victims Unit" (high hit count / sort check)

Used to confirm result sorting behaves correctly on a large result
set, not just the low-count "benson" tests above.

### Freeview + OpenTV

| Image/Binary | Search type | Result count | 28.2E count | DVB-T count |
|---|---|---|---|---|
| ATV 8.0 stock | 2 (partial) | 88 | 56 | 32 |
| wxbet modified | 2 (partial) | 89 | 56 | 33 |
| ATV 8.0 stock | 5 (partialdes) | 0 | 0 | 0 |
| wxbet modified | 5 (partialdes) | 0 | 0 | 0 |

PR #3877, opened 2026-08-29
[SoftCSA] Persist CSA-ALT detection and enforce buffered startup gates

ManuGH:
## Problem

SoftCSA startup had several ordering and persistence issues:

- Positive CSA-ALT detection was cached only for the lifetime of the Enigma2 process. After a GUI restart, an already known service had to be classified again.
- `eventNewProgramInfo` could call `updatePids(true)` while only the recorder was active. This created the decoder before the first-CW and pre-buffer gates had completed.
- The configured pre-buffer was a blind wall-clock delay. Depending on CW timing, PES cadence, and DVR queue pressure, the decoder could start with too little clear audio/video data or wait until the idle queue filled.

## Changes

- Persist positive CSA-ALT detections by DVB triplet in `${sysconfdir}/enigma2/softcsa.cache`.
- Load the cache before rejecting the initial `0xFFFF` ECM PID placeholder, allowing known services to select SoftCSA without waiting for a second classification.
- Continue monitoring the first valid ECM to validate cached entries. Stale entries are removed automatically and playback returns to hardware descrambling.
- Write the persistent cache only when its serialized positive classification or nonzero service ID changes. Failed writes remain dirty and are retried at the next save opportunity.
- Preserve the first-CW and pre-buffer gates on PMT updates by passing `m_decoder_started`, rather than `m_running`, to `updatePids()`.
- Track clear packets for the selected video and audio PIDs after software descrambling, extract PES PTS values, and derive actual buffered media duration.
- Start the decoder when both selected streams have reached the configured `bufferTime`. The last observed PES interval is included because a PTS marks the beginning of a block; a bounded 50 ms tolerance covers timestamp/timer quantization.
- Retain a `bufferTime + 500 ms` safety deadline for missing or unusable timestamps.
- Disable readiness tracking as soon as the decoder starts, leaving no steady-state packet-scanning or locking overhead.

Only positive detections are persisted. Writes use a temporary file plus `rename()`. FTA, process-local negative detections, and other CA systems retain their existing hardware path. CI activation suppression, timeout fallback, and the existing Aggressive-mode cache bypass remain in place.

The PR deliberately retains the existing hardware-decoder startup order for unknown services. Instrumented control runs did not demonstrate a visible hardware-to-software dropout, so waiting for local ECM classification did not justify an additional hardware-start gate.

## Test

- Rebased on current `master` (`414e49e707`) and completed a clean ARMv7 `vuuno4k` Docker build.
- Deployed the binary and matching generated Python wrapper to an openATV 8.0 `vuuno4k`.
- On an anonymized encrypted CSA-ALT test service with `bufferTime=1000`, clear A/V readiness completed after 965 ms. SoftDecoder PLAY followed, and confirmed video and audio PTS were present 121 ms after the decoder start request.
- Left the SoftCSA service running continuously for five minutes. Twenty video-only grabs taken at 15-second intervals were all different; ECM data kept updating and the Enigma2 process remained unchanged.
- The five-minute log contained no traceback, fatal error, TS write error, recorder error, buffer overrun/underrun, decoder-start failure, or softcam-proxy disconnect.
- Repeated startup and tune cycles for an anonymized non-CSA-ALT encrypted service. Its hardware decoder retained the normal immediate startup order and produced changing video output in both complete control runs.
- Starting without a cache recreated the positive classification. After a subsequent GUI restart, cached validation, ECM handling, and first-CW delivery left the cache inode, modification time, size, and content unchanged.
- Tuning an unrelated negative VideoGuard service also left the positive-only cache unchanged.
- Cached positive classifications select SoftCSA directly. Negative classifications remain process-local and retain their immediate hardware-decoder path.

This complements the pre-buffer work merged in #3720 by ensuring PMT updates cannot bypass the gate and by basing decoder startup on clear media readiness instead of elapsed time alone.

### DIFF
--- a/lib/dvb/csasession.cpp
+++ b/lib/dvb/csasession.cpp
@@ -2,8 +2,14 @@
 #include <lib/dvb/csaengine.h>
 #include <lib/dvb/cahandler.h>
 #include <lib/dvb/cwhandler.h>
+#include <lib/base/eenv.h>
 #include <lib/base/eerror.h>
 #include <lib/base/esimpleconfig.h>
+
+#include <cstdio>
+#include <fstream>
+#include <iomanip>
+#include <sstream>
 
 #ifdef DREAMNEXTGEN
 #include <lib/dvb/alsa.h>
@@ -23,6 +29,89 @@ struct ServiceCsaInfo {
 	bool serviceId_valid; // true if serviceId has been seen
 };
 static std::map<uint64_t, ServiceCsaInfo> s_csa_cache;
+static bool s_persistent_csa_cache_loaded = false;
+static bool s_persistent_csa_cache_dirty = false;
+
+static std::string persistentCsaCachePath()
+{
+	return eEnv::resolve("${sysconfdir}/enigma2/softcsa.cache");
+}
+
+// Keep positive CSA-ALT detections across GUI restarts so known services can
+// select SoftCSA immediately without repeating unnecessary HW decoder setup.
+static void loadPersistentCsaCache()
+{
+	if (s_persistent_csa_cache_loaded)
+		return;
+
+	s_persistent_csa_cache_loaded = true;
+	std::ifstream in(persistentCsaCachePath());
+	if (!in.good())
+		return;
+
+	unsigned int loaded = 0;
+	std::string line;
+	while (std::getline(in, line))
+	{
+		if (line.empty() || line[0] == '#')
+			continue;
+
+		unsigned long long key = 0;
+		unsigned int ecm_mode = 0;
+		unsigned int service_id = 0;
+		std::istringstream fields(line);
+		if (!(fields >> std::hex >> key >> ecm_mode >> service_id) || ecm_mode > 0x0F)
+			continue;
+
+		ServiceCsaInfo &info = s_csa_cache[static_cast<uint64_t>(key)];
+		info.is_csa_alt = true;
+		info.ecm_mode = static_cast<uint8_t>(ecm_mode);
+		info.valid = true;
+		info.serviceId = service_id;
+		info.serviceId_valid = service_id != 0;
+		++loaded;
+	}
+
+	eDebug("[eDVBCSASession] Loaded %u persistent CSA-ALT service(s)", loaded);
+}
+
+static void flushPersistentCsaCache()
+{
+	if (!s_persistent_csa_cache_dirty)
+		return;
+
+	const std::string path = persistentCsaCachePath();
+	const std::string temporary_path = path + ".tmp";
+	std::ofstream out(temporary_path, std::ios::trunc);
+	if (!out.good())
+	{
+		eWarning("[eDVBCSASession] Failed to write persistent cache %s", temporary_path.c_str());
+		return;
+	}
+
+	out << "# Enigma2 SoftCSA CSA-ALT cache v1\n";
+	for (const auto &entry : s_csa_cache)
+	{
+		const ServiceCsaInfo &info = entry.second;
+		if (!info.valid || !info.is_csa_alt)
+			continue;
+
+		out << std::hex << std::setfill('0')
+			<< std::setw(16) << entry.first << ' '
+			<< std::setw(2) << static_cast<unsigned int>(info.ecm_mode) << ' '
+			<< std::setw(8) << (info.serviceId_valid ? info.serviceId : 0) << '\n';
+	}
+	out.close();
+
+	if (!out.good() || std::rename(temporary_path.c_str(), path.c_str()) != 0)
+	{
+		eWarning("[eDVBCSASession] Failed to replace persistent cache %s", path.c_str());
+		std::remove(temporary_path.c_str());
+		return;
+	}
+
+	s_persistent_csa_cache_dirty = false;
+}
 
 // Helper: Check if CAID is VideoGuard
 static bool caid_is_videoguard(uint16_t caid)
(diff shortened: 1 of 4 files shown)


WXbet:
Can you document the dropout with a log? The SoftDecoder only starts upon `firstCwReceived`, and by design, the local ECM analysis completes before the CW arrives from OScam. With a CSA-ALT service, the hardware decoder hasn't even delivered an image yet at the time of the handover—so what exactly is supposed to drop out?

ManuGH:
You're right. I added monotonic logging around ECM analysis, hardware-decoder release, the first CW, SoftDecoder PLAY, and the first output PTS, then tested an uncached CSA-ALT tune on a `vuuno4k`:

```text
20:10:41.9529 Deferring HW decoder until ECM decision (timeout=1500ms)
20:10:42.0553 [SoftCSA-HANDOVER] t=4829214ms ECM analyzed: caid=0x098C ecm_mode=0x04 CSA-ALT=1
20:10:42.0576 [SoftCSA-HANDOVER] t=4829217ms before HW release: present=0
20:10:42.1168 [SoftCSA-HANDOVER] t=4829276ms first CW received
20:10:43.0940 Adaptive pre-buffer ready after 977ms (video=1240ms/3382 packets, audio=960ms/247 packets)
20:10:43.1805 [SoftCSA-HANDOVER] t=4830339ms SoftDecoder PLAY vpid=02ff vpidtype=1
20:10:43.2127 [SoftCSA-HANDOVER] t=4830372ms SoftDecoder first video PTS after 119ms
20:10:43.2128 [SoftCSA-HANDOVER] t=4830372ms SoftDecoder first audio PTS after 119ms
```

So there is no hardware picture or audio to drop at that point, and I cannot document the claimed dropout. The original problem statement was incorrect. I have updated the PR title, description, and first commit wording accordingly.

The observed reasons for the remaining changes are the PMT path calling `updatePids(true)` while only the recorder is running, which can bypass the first-CW/pre-buffer gates, and replacing the wall-clock pre-buffer with measured clear A/V readiness. The cache and ECM decision gate avoid unnecessary hardware-decoder creation and make classification persistent; I no longer describe that as fixing a visible dropout.

ManuGH:
@WXbet You were right. I repeated the test on current `master` with monotonic diagnostics so the existing path was measured directly.

The uncached control run produced this relative timeline:

```text
+   0 ms  HW decoder PLAY
+ 170 ms  ECM analyzed: CSA-ALT=1
+ 170 ms  HW decoder release begins
+2104 ms  first CW received
+3343 ms  SoftDecoder PLAY
+3385 ms  first confirmed SoftDecoder video and audio PTS
```

There was no hardware video event between HW PLAY and release. The decoder object existed briefly, but there was no decoded-frame event showing that a picture had been displayed. I therefore cannot document the previously claimed visible dropout.

Follow-up: I removed the ECM decision gate from the PR. Unknown and non-CSA-ALT services now retain the existing immediate hardware-decoder startup order. Additional hardware-path startup tests produced live output both with and without the experimental gate, but without a demonstrated visible benefit the extra normal-service startup path was not justified.

The PR is now limited to persistent positive classification, preventing PMT updates from bypassing the first-CW/pre-buffer gates, and starting SoftDecoder from measured clear A/V readiness.

WXbet:
Dropping the gate was the right way, and thanks for measuring on plain `master`.

That control run also answers what's left:

```text
+ 170 ms  ECM analyzed: CSA-ALT=1
+2104 ms  first CW received
+3343 ms  SoftDecoder PLAY
```

Classification is done 1934 ms before the CW, and the decoder start is gated on
`firstCwReceived()`. Persisting it moves +170 ms to +0 ms and changes nothing, the start
still waits for +2104 ms. And as you confirmed, there was no hardware video event, so the
decoder object the cache avoids creating never showed a picture anyway.

Against that zero gain:

- The persisted `serviceId` is OScam-internal and stale after a reboot or an OScam restart,
  but `setActive()` pre-registers with it. Your new `service_id_changed` check cannot
  correct it: it sits in `onCwReceived()` inside `if (!m_cw_handler_registered)`, which the
  pre-registration has already set. The `else if` branch keeps CWs flowing, so nothing
  breaks, but the wrong ID stays in the file permanently and every zap registers the
  engine twice.
- A stale positive entry activates SoftCSA and is only corrected when the real ECM is
  parsed, via `setActive(false)` → SoftDecoder stop → hardware decoder rebuild. That
  normally lands in the same no-picture window as everything else here. But if keys are
  already available at start (`hasKeys()`, e.g. a recording of the same service is running)
  the SoftDecoder is showing video by then and the rebuild is a real dropout.
- No cap, no aging, no `fsync()` before the `rename()`.
- On the cached-positive path `m_ecm_analyzed` stays false, so `startECMMonitor()` rebuilds
  the section reader on every PMT event until the first ECM arrives.

## The pre-buffer is already a setting
Your earlier run had the readiness gate releasing at 977 ms against a 1000 ms target,
23 ms off the existing wall-clock timer, while the new `bufferTime + 500 ms` deadline can
make it 500 ms slower. One step on the existing selector moves more than the whole
mechanism does. Both options are under **SoftCSA Settings**, setup level 2 (Expert):

**Decoder start timeout** (`config.softcsa.waitForDataTimeout`, default *Disabled*)
I think this is the piece you're missing. At *Disabled* the CW wait is skipped and the
decoder starts on null packets, because the engine has no key yet. It is **a ceiling, not
a delay**: `onFirstCwReceived()` stops the timer immediately and `hasKeys()` short-circuits
it on a fast switch, so a CW at +2104 ms starts the decoder at +2104 ms whether you set
500 ms or 2000 ms. Setting it generously is free, put it above your worst CW latency.

**Decoder start delay** (`config.softcsa.bufferTime`, default *Disabled*)
this one *is* paid in full on every zap, so tune it downward. With the timeout enabled,
the window it covers holds real media by construction instead of a mix of nulls and
media. That is the property your readiness gate measures at runtime, given structurally
and for free.

## On testing scope
One box isn't enough to justify a change here, whichever box it is. SoftCSA runs through
decoder handover, the DVR loopback and the descrambler slots; all driver territory, and
the drivers differ a lot between chipsets. The code already carries several per-platform
exceptions in exactly these places.

The same descrambling path is also driven by live TV, timeshift, recording, streaming and
transcoding, each with its own timing and its own consumer. Your list covers live TV on
one receiver, but nothing on the other 4 paths.

## The one-liner
`updatePids(m_decoder_started)` is correct, `m_running` only means the recorder is up. But
on a stock configuration `setupRecorder()` already runs `startDecoder()` internally (both
options default to *Disabled*), so `m_decoder_started` is true before `m_running` is set
and the two are identical. Worth folding in next time that file is touched; not worth a PR
on its own.

Thanks for the investigation and please try your use case with the existing SoftCSA options first.

PR #3878, opened 2026-08-29
dvb: ensure disabled tuners (configMode=nothing) never participate in tuner selection

catalinii:
* NimManager: clear delivery-system whitelist for disabled single-type tuners (configMode=nothing)

Tuners A-H are DVB-C BCM3158 with dvbc.configMode=nothing and must not advertise DVB-C. Previously SecConfigure only cleared the whitelist for multi-type slots (continue), while single-type did setFrontendType( slot.getType()) unconditionally, so a disabled BCM3158 still reported SYS_DVBC_ANNEX_A via m_delsys and could be scored 2 before the m_enabled check. Now single-type checks configMode and calls setFrontendType( 'dummy', False) for nothing, so supportsDeliverySystem() fails and isCompatibleWith() returns 0. Tuner J (SAT>IP VTUNER slot 9, DVB-C) remains the sole DVB-C candidate; A-H never participate in allocateFrontend.

Tested on vuuno4kse (8x BCM3158 + 2x SAT>IP VTUNER, frontend_priority=9, A-H nothing, J enabled).

* NimManager: guard IndexError when VTUNER/minisatip creates extra NIM slots beyond config.Nims

On vuuno4kse with local minisatip -a 0:0:8, NIM sockets 10/11 (SAT>IP VTUNER frontend 10/11) appear while config.Nims only has 0-9. InitNimManager accessed config.Nims[slot_id] without bounds check at two loops (2501, 2616), raising IndexError: list index out of range and aborting StartEnigma (release cached channel timer). Add if slot_id >= len(config.Nims): continue.

---------

### DIFF
--- a/lib/python/Components/NimManager.py
+++ b/lib/python/Components/NimManager.py
@@ -359,7 +359,12 @@ class SecConfigure:
 							continue
 						eDVBResourceManager.getInstance().setFrontendType(slot.frontend_id, FeType, True)
 				else:
-					eDVBResourceManager.getInstance().setFrontendType(slot.frontend_id, slot.getType())
+					# For single-type tuners, respect configMode=nothing: disabled tuners (A-H with
+					# dvbc.configMode=nothing) must not advertise a delivery system, otherwise
+					# they could still be considered for tuning despite m_enabled=False.
+					slot_type = slot.getType()
+					if slot_type not in ("DVB-C", "DVB-C2") or config.Nims[slot.slot].dvbc.configMode.value != "nothing":
+						eDVBResourceManager.getInstance().setFrontendType(slot.frontend_id, "dummy", False)
 		print("[NimManager] sec config completed")
 
 	def updateAdvanced(self, sec, slotid):
@@ -2481,6 +2486,8 @@ def InitNimManager(nimmgr, update_slots=None):
 	empty_slots = 0
 	for slot in nimmgr.nim_slots:
 		slot_id = slot.slot
+		if slot_id >= len(config.Nims):
+			continue
 		nim = config.Nims[slot_id]
 		nim.force_legacy_signal_stats = ConfigYesNo(default=False)
 		if slot.canBeCompatible("DVB-S"):
@@ -2596,6 +2603,8 @@ def InitNimManager(nimmgr, update_slots=None):
 	empty_slots = 0
 	for slot in nimmgr.nim_slots:
 		slot_id = slot.slot
+		if slot_id >= len(config.Nims):
+			continue
 		nim = config.Nims[slot_id]
 		addMultiType = False
 		try:


atvcaptain:
Thank you for the patch. We understand that it addresses two separate problems:

1. Built-in DVB-C FBC tuners configured as `nothing connected` should not be selected for tuning.
2. An `IndexError` occurs if an additional VTUNER is registered after Enigma2 has already initialized `config.Nims`.

For the first point, Enigma2 already passes the configured state to the C++ frontend code through `setFrontendSlotInformations()`. A tuner configured as `nothing` receives `m_enabled = false`, and `isCompatibleWith()` returns `0` for a disabled frontend before checking its advertised delivery systems.

The tuner remains visible in `/proc/bus/nim_sockets` because that list is provided by the driver. This does not mean that Enigma2 can select it for tuning.

The proposed `setFrontendType()` condition also appears to be inverted. It would set active DVB-C tuners and every non-DVB-C single-type tuner to `dummy`, while not changing a disabled DVB-C tuner. We therefore do not plan to include that part.

Our understanding of the `IndexError` is:

- Enigma2 starts with, for example, 17 detected slots.
- It creates 17 entries in `config.Nims`.
- The SAT>IP client finishes registering another VTUNER later.
- A subsequent enumeration finds 18 slots.
- Accessing `config.Nims[17]` then raises an `IndexError`.

Enigma2 does not currently support adding a completely new tuner configuration at runtime. Adding or changing VTUNERs already requires a GUI restart. Instead of implementing incomplete tuner hotplug support or adding guards to only some of the affected loops, we propose deferring completely new slot indices until the next GUI restart:

```diff
diff --git a/lib/python/Components/NimManager.py b/lib/python/Components/NimManager.py
@@
 		nimfile.close()
+		# The tuner configuration is created once from the slots present during
+		# Enigma2 startup. A VTUNER may finish registering later, but Enigma2
+		# does not support adding a new tuner configuration at runtime. Keep
+		# updates to already known (including previously empty) slots, and defer
+		# completely new slot indices until the next GUI restart.
+		try:
+			configured_slots = len(config.Nims)
+		except AttributeError:
+			configured_slots = None
+		if configured_slots is not None:
+			late_slots = sorted(slot_id for slot_id in entries if slot_id >= configured_slots)
+			if late_slots:
+				print(f"[NimManager] New NIM slot(s) {late_slots} detected after tuner configuration initialization; a GUI restart is required before they can be used.")
+				entries = {slot_id: entry for slot_id, entry in entries.items() if slot_id < configured_slots}
 		self.number_of_slots = len(list(entries.keys()))
```

This keeps updates for already known slots, including previously empty VTUNER slots, but prevents `nim_slots` from becoming longer than `config.Nims` during the running session.

We reproduced and tested the problem on a Vu+ Duo 4K Lite:

```text
Before enumeration: nim_slots=17, config.Nims=17
/proc/bus/nim_sockets: 18 slots
After enumeration: nim_slots=17, config.Nims=17
InitNimManager completed without an exception
```

The additional VTUNER is detected normally after restarting Enigma2.

Please confirm whether this matches the problem you intended to solve. In particular, are the disabled DVB-C FBC tuners actually being allocated for channel tuning, or are they only still visible in `/proc/bus/nim_sockets`? If they are being allocated despite being configured as `nothing`, please provide an Enigma2 debug log showing the frontend enable state and tuner allocation.

atvcaptain:
https://github.com/openatv/enigma2/commit/a4c0751a689ec5a24d9060615d5fbbdf364f3b71

PR #3895, opened 2026-09-07
[Screen] fix SummaryScreen scaling problem

xcentaurix:
display skins should not be scaled because lcd dimensions are fixed.
however, applySkin does not differentiate between osd and lcd skins and thus lcd summary screens are scaled as well.
this code change differentiates between osd and lcd skins and fixes the problem.

### DIFF
--- a/lib/python/Screens/Screen.py
+++ b/lib/python/Screens/Screen.py
@@ -3,7 +3,7 @@ from types import CodeType
 
 from enigma import eRCInput, eStack, eTimer, eWindow, getDesktop
 
-from skin import GUI_SKIN_ID, applyAllAttributes, menus, readSkin, screens, setups
+from skin import DISPLAY_SKIN_ID, GUI_SKIN_ID, applyAllAttributes, menus, readSkin, screens, setups
 from Components.ActionMap import HelpableActionMap
 from Components.config import config
 from Components.GUIComponent import GUIComponent
@@ -289,7 +289,8 @@ class Screen(dict):
 				method()
 
 	def applySkin(self):
-		bounds = (getDesktop(GUI_SKIN_ID).size().width(), getDesktop(GUI_SKIN_ID).size().height())
+		skinId = DISPLAY_SKIN_ID if isinstance(self, ScreenSummary) else GUI_SKIN_ID
+		bounds = (getDesktop(skinId).size().width(), getDesktop(skinId).size().height())
 		resolution = bounds
 		zPosition = 0
 		for (key, value) in self.skinAttributes:


jbleyel:
I'm a little bit confused because scaling is only active if you have a resolution attribute in the skin.
Do you have display skins with resolution attribute? Why?

xcentaurix:
yes, for lcd only skins this works without resolution tag.
but when you have a plugin skin with osd screens and a summary screen the summary screen is always scaled even without resolution tag.

jbleyel:
> yes, for lcd only skins this works without resolution tag. but when you have a plugin skin with osd screens and a summary screen the summary screen is always scaled even without resolution tag.

Can you show me an example?

xcentaurix:
```
<screen name="BufferingProgressSummary" position="fill" resolution="396,240">
	<ePixmap pixmap="~/../images/display_bg.png" position="0,0" size="400,240" zPosition="-1"/>
	<widget name="lcd_pic" position="150,70" size="100,100" transparent="1"/>
</screen>
```
again, it doesnt make a difference with or without resolution tag.

jbleyel:
> ```
> <screen name="BufferingProgressSummary" position="fill" resolution="396,240">
> 	<ePixmap pixmap="~/../images/display_bg.png" position="0,0" size="400,240" zPosition="-1"/>
> 	<widget name="lcd_pic" position="150,70" size="100,100" transparent="1"/>
> </screen>
> ```
> 
> again, it doesnt make a difference with or without resolution tag.

Sorry. I mean an example plugin where you have issues.

xcentaurix:
https://github.com/OpenCockpit/MediathekCockpit

jbleyel:
> https://github.com/OpenCockpit/MediathekCockpit

this is not correct :

`<screen backgroundColor="backgroundMedium" flags="wfNoBorder" name="CockpitPlayerSummary" position="fill" resolution="396,240">`

xcentaurix:
yes, it just slipped in recently and i will change it back. however, this doesnt cause the scaling problem.
here is claude's explanation:
```
# Screen.applySkin() always derives its scale/resolution reference from
# getDesktop(GUI_SKIN_ID) (the 1920x1080 GUI desktop), even for a screen
# bound to a different desktop (e.g. the LCD summary desktop) via
# setDesktop().
```

i will create an lcd screenshot  later to show what's displayed.

xcentaurix:
this is the lcd screenshot without the fix in this pr:
lcd_bug

and this is the lcd screenshot with the fix (same as openvix without a change):
lcd_with_fix

jbleyel:
I think your screen must be this:

`<screen backgroundColor="backgroundMedium" flags="wfNoBorder" name="CockpitPlayerSummary" position="0,0" size="396,240">`

But this will only work for boxes with a 400* 240 display.
So you don't wanna support other boxes?

xcentaurix:
no, it's optimized for this size.

jbleyel:
> no, it's optimized for this size.

And what about the other display resolutions especially 7 Segment?

I think you have only 2 options.
A) use an existing Summary Screen
B) add your Summary Screen  to all display skins.

Warning!!
B must be done on all images and here -> https://github.com/oe-alliance/enigma2-display-skins

xcentaurix:
well, this is a different discussion. the scaling bug should be fixed anyway because it doesnt make sense to use the osd size for any lcd scaling.

jbleyel:
https://github.com/openatv/enigma2/pull/3901

PR #3911, opened 2026-09-17
Consolidate action key constants into an ActionKeys class and make re…

jbleyel:
…adonly styling skin-configurable

Replace scattered ACTIONKEY_* module constants with an ActionKeys class and match/case dispatch across ConfigElement subclasses in config.py, and mark read-only config entries with a distinct text color in getMulti(). Update all call sites (ConfigList.py, Wizard.py, Ci.py, PositionerSetup) to reference ActionKeys.ACTIONKEY_* instead of the old bare module constants.

Move read-only enforcement from individual ConfigElement.handleKey() implementations to the UI layer: ConfigListScreen now disables the relevant action maps for read-only entries, and Wizard gained a handleConfigKey() helper doing the same isReadOnly() check for its direct handleKey() calls, since Wizard does not use ConfigListScreen.

Make the read-only text color skinnable via a new readOnlyColor attribute on <windowstyle><configList>, applied through Components.config.setReadOnlyColor() and reset to the default by reloadSkins().


IanSav (approved):
The config.py and ConfigList.py modules are effectively the same as I coded and we tested.  The other changes are a simple and logical progressions of these optimisations.  I have patched my system with these changes and can't seem to fault the new code.  I feel confident that this branch is good to go.

